PR #9, opened 2019-01-03
BAH-625 Add self privilege pr

nehagupta2123:
https://bahmni.atlassian.net/browse/BAH-625?jql=

### DIFF
--- a/api/src/main/java/org/openmrs/module/appointments/service/AppointmentsService.java
+++ b/api/src/main/java/org/openmrs/module/appointments/service/AppointmentsService.java
@@ -12,46 +12,50 @@ import org.springframework.transaction.annotation.Transactional;
 import java.util.Date;
 import java.util.List;
 
+import static org.openmrs.module.appointments.constants.PrivilegeConstants.MANAGE_APPOINTMENTS;
+import static org.openmrs.module.appointments.constants.PrivilegeConstants.MANAGE_OWN_APPOINTMENTS;
+import static org.openmrs.module.appointments.constants.PrivilegeConstants.VIEW_APPOINTMENTS;
+
 public interface AppointmentsService {
 
     @Transactional
-    @Authorized({"Manage Appointments", "Manage Own Appointments"})
+    @Authorized({MANAGE_APPOINTMENTS, MANAGE_OWN_APPOINTMENTS})
     Appointment validateAndSave(Appointment appointment);
 
     @Transactional
-    @Authorized({"View Appointments"})
+    @Authorized({VIEW_APPOINTMENTS})
     List<Appointment> getAllAppointments(Date forDate);
 
     @Transactional
-    @Authorized({"View Appointments"})
+    @Authorized({VIEW_APPOINTMENTS})
     List<Appointment> search(Appointment appointment);
 
     @Transactional
-	@Authorized({"View Appointments"})
+    @Authorized({VIEW_APPOINTMENTS})
     List<Appointment> getAllFutureAppointmentsForService(AppointmentServiceDefinition appointmentServiceDefinition);
 
     @Transactional
-	@Authorized({"View Appointments"})
+    @Authorized({VIEW_APPOINTMENTS})
     List<Appointment> getAllFutureAppointmentsForServiceType(AppointmentServiceType appointmentServiceType);
 
     @Transactional
-	@Authorized({"View Appointments"})
+    @Authorized({VIEW_APPOINTMENTS})
     List<Appointment> getAppointmentsForService(AppointmentServiceDefinition appointmentServiceDefinition, Date startDate, Date endDate, List<AppointmentStatus> appointmentStatusList);
 
     @Transactional
-	@Authorized({"View Appointments"})
+    @Authorized({VIEW_APPOINTMENTS})
     Appointment getAppointmentByUuid(String uuid);
 
     @Transactional
-    @Authorized({"Manage Appointments", "Manage Own Appointments"})
+    @Authorized({MANAGE_APPOINTMENTS, MANAGE_OWN_APPOINTMENTS})
     void changeStatus(Appointment appointment, String status, Date onDate);
 
     @Transactional
-	@Authorized({"View Appointments"})
+    @Authorized({VIEW_APPOINTMENTS})
     List<Appointment> getAllAppointmentsInDateRange(Date startDate, Date endDate);
 
     @Transactional
-    @Authorized({"Manage Appointments", "Manage Own Appointments"})
+    @Authorized({MANAGE_APPOINTMENTS, MANAGE_OWN_APPOINTMENTS})
     void undoStatusChange(Appointment appointment);
 
     @Transactional
--- a/api/src/main/java/org/openmrs/module/appointments/service/AppointmentsService.java
+++ b/api/src/main/java/org/openmrs/module/appointments/service/AppointmentsService.java
@@ -15,7 +15,7 @@ import java.util.List;
 public interface AppointmentsService {
 
     @Transactional
-    @Authorized({"Manage Appointments", "Manage Self Appointments"})
+    @Authorized({"Manage Appointments", "Manage Own Appointments"})
     Appointment validateAndSave(Appointment appointment);
 
     @Transactional
@@ -43,7 +43,7 @@ public interface AppointmentsService {
     Appointment getAppointmentByUuid(String uuid);
 
     @Transactional
-    @Authorized({"Manage Appointments", "Manage Self Appointments"})
+    @Authorized({"Manage Appointments", "Manage Own Appointments"})
     void changeStatus(Appointment appointment, String status, Date onDate);
 
     @Transactional
@@ -51,7 +51,7 @@ public interface AppointmentsService {
     List<Appointment> getAllAppointmentsInDateRange(Date startDate, Date endDate);
 
     @Transactional
-    @Authorized({"Manage Appointments", "Manage Self Appointments"})
+    @Authorized({"Manage Appointments", "Manage Own Appointments"})
     void undoStatusChange(Appointment appointment);
 
     @Transactional
--- a/api/src/main/java/org/openmrs/module/appointments/service/AppointmentsService.java
+++ b/api/src/main/java/org/openmrs/module/appointments/service/AppointmentsService.java
@@ -15,15 +15,15 @@ import java.util.List;
 public interface AppointmentsService {
 
     @Transactional
-	@Authorized({"Manage Appointments"})
+    @Authorized({"Manage Appointments", "Manage Self Appointments"})
     Appointment validateAndSave(Appointment appointment);
 
     @Transactional
-	@Authorized({"View Appointments"})
+    @Authorized({"View Appointments"})
     List<Appointment> getAllAppointments(Date forDate);
 
     @Transactional
-	@Authorized({"View Appointments"})
+    @Authorized({"View Appointments"})
     List<Appointment> search(Appointment appointment);
 
     @Transactional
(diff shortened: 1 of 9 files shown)


api/src/main/java/org/openmrs/module/appointments/service/AppointmentsService.java
```diff
@@ -12,47 +12,51 @@
...
 
+import static org.openmrs.module.appointments.constants.PrivilegeConstants.MANAGE_APPOINTMENTS;
+import static org.openmrs.module.appointments.constants.PrivilegeConstants.MANAGE_OWN_APPOINTMENTS;
+import static org.openmrs.module.appointments.constants.PrivilegeConstants.VIEW_APPOINTMENTS;
+
 public interface AppointmentsService {
 
     @Transactional
-	@Authorized({"Manage Appointments"})
+    @Authorized({MANAGE_APPOINTMENTS, MANAGE_OWN_APPOINTMENTS})
```
ivange94: My understanding is that if I'm a provider, I should be able to only modify my own appointments. But does this prevent that from happening? I mean if the authorized user is has only the MANAGE_OWN_APPOINTMENTS, they are still able to call this method even when the appointment they are editing is not theirs.

api/src/main/java/org/openmrs/module/appointments/service/impl/AppointmentsServiceImpl.java
```diff
@@ -21,11 +23,18 @@
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Set;
 import java.util.stream.Collectors;
 
+import static java.util.Objects.isNull;
+
+import static org.openmrs.module.appointments.constants.PrivilegeConstants.MANAGE_APPOINTMENTS;
+import static org.openmrs.module.appointments.model.AppointmentProviderResponse.ACCEPTED;
```
ivange94: I don't think this ever used.

api/src/test/java/org/openmrs/module/appointments/service/AppointmentsServiceTest.java
```diff
@@ -67,6 +72,24 @@ public void init() throws Exception {
         executeDataSet("appointmentTestData.xml");
     }
 
+    @Test
+    public void shouldSaveAppointmentsOnlyIfUserHasManageOwnPrivilege() throws Exception {
+        Context.authenticate(manageOwnUser, manageOwnUserPassword);
+        Appointment appointment = getSampleAppointment();
+        assertNotNull(appointmentsService.validateAndSave(appointment));
+    }
+
```
ivange94: Can you add a test that shows that if a provider is editing an appointment they are not a provider for it will fail?

api/src/main/java/org/openmrs/module/appointments/service/AppointmentsService.java
sai789: Yes, true. It's a common method to create and edit an appointment for any(both manage and own) privileged providers. In that method's implementation, we are validating the users privileges to take decision on further action. We did not create a separate method for this as that would have been two methods for each action with duplicate code.

api/src/main/java/org/openmrs/module/appointments/service/impl/AppointmentsServiceImpl.java
sai789: Agreed. We will remove it.

api/src/test/java/org/openmrs/module/appointments/service/AppointmentsServiceTest.java
sai789: Sure . Will add those tests

sai789: The tests are already there for these scenarios in AppointmentsServiceImplTest class. Is there any specific reason we need to mention the tests at this level.

api/src/main/java/org/openmrs/module/appointments/service/impl/AppointmentsServiceImpl.java
ivange94: No you don't have to mention them twice. I probably missed them in my first review. I'll have a look again.

api/src/test/java/org/openmrs/module/appointments/service/AppointmentsServiceTest.java
ivange94: no you don't have to mention them twice. I probably missed them in my initial review. Will look again.

api/src/test/java/org/openmrs/module/appointments/service/AppointmentsServiceTest.java
```diff
@@ -102,135 +125,149 @@ public void shouldSaveAppointmentsOnlyIfUserHasManagePrivilege() throws Exceptio
...
         appointmentsService.undoStatusChange(appointment);
     }
 
+    @Test(expected = APIException.class)
+    public void shouldBeAbleToUndoStatusChangeIfUserHasManageOwnPrivilege() {
+        Context.authenticate(manageOwnUser, manageOwnUserPassword);
+        Appointment appointment = new Appointment();
+        appointment.setId(1);
+        appointmentsService.undoStatusChange(appointment);
+    }
```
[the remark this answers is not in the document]
nehagupta2123: The API exception is due to the reason that there are no status change actions to undo. If the user does not have the privilege to perform undo operation the exception thrown would be authentication exception. The mock setup for appointment is not done at this class because this is just to check whether the user with appropriate privileges is able to hit the method.

api/src/main/java/org/openmrs/module/appointments/service/impl/AppointmentsServiceImpl.java
```diff
@@ -52,8 +60,28 @@ public void setAppointmentAuditDao(AppointmentAuditDao appointmentAuditDao) {
         this.appointmentAuditDao = appointmentAuditDao;
     }
 
+    private boolean validateIfUserHasSelfOrAllAppointmentsAccess(Appointment appointment) {
+        return Context.hasPrivilege(MANAGE_APPOINTMENTS) ||
+                isAppointmentForNoProvider(appointment) ||
+                isCurrentUserSamePersonAsOneOfTheAppointmentProviders(appointment.getProviders());
+    }
+
+    private boolean isAppointmentForNoProvider(Appointment appointment) {
```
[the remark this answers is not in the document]
nehagupta2123: Done.

api/src/test/java/org/openmrs/module/appointments/service/impl/AppointmentsServiceImplTest.java
```diff
@@ -366,4 +408,39 @@ public void shouldThrowExceptionWhenThereIsNoPriorStatusChangeExists() throws Pa
         verify(appointmentDao, times(0)).save(appointment);
     }
 
+    @Test(expected = APIAuthenticationException.class)
+    public void shouldThrowExceptionOnAppointmentSaveIfUserHasOnlyOwnPrivilegeAndProviderAndUserIsNotTheSamePerson() {
+        setupForOwnPrivilegeAccess(exceptionCode);
+
+        appointmentsService.validateAndSave(appointment);
```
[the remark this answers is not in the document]
nehagupta2123: Done. Please review it once if that is what you suggested.

api/src/test/java/org/openmrs/module/appointments/service/AppointmentsServiceTest.java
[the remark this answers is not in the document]
nehagupta2123: We're checking if the user has appropriate privileges can hit the method. But if the method is hit without setting the appointment context, it will throw an exception.

api/src/test/java/org/openmrs/module/appointments/service/impl/AppointmentsServiceImplTest.java
```diff
@@ -331,13 +371,14 @@ public void shouldUndoStatusChange() throws ParseException {
         when(appointmentAuditDao.getPriorStatusChangeEvent(appointment)).thenReturn(appointmentAudit);
         appointmentsService.undoStatusChange(appointment);
         verify(appointmentAuditDao, times(1)).getPriorStatusChangeEvent(appointment);
-        ArgumentCaptor<Appointment> captor = ArgumentCaptor.forClass(Appointment.class);;
+        ArgumentCaptor<Appointment> captor = ArgumentCaptor.forClass(Appointment.class);
+        ;
```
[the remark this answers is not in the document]
nehagupta2123: done

api/src/main/java/org/openmrs/module/appointments/service/impl/AppointmentsServiceImpl.java
```diff
@@ -52,8 +60,28 @@ public void setAppointmentAuditDao(AppointmentAuditDao appointmentAuditDao) {
         this.appointmentAuditDao = appointmentAuditDao;
     }
 
+    private boolean validateIfUserHasSelfOrAllAppointmentsAccess(Appointment appointment) {
+        return Context.hasPrivilege(MANAGE_APPOINTMENTS) ||
+                isAppointmentNotAssignedToAnyProvider(appointment) ||
+                isCurrentUserSamePersonAsOneOfTheAppointmentProviders(appointment.getProviders());
+    }
+
+    private boolean isAppointmentNotAssignedToAnyProvider(Appointment appointment) {
```
angshu: Minor changes required here. In multiple provider scenario, an appointment can have multiple providers, but their status might be all rejected. That needs to be considered as well. For example: "appointment.getProviders().size() == EMPTY_SET_SIZE" is not enough, since there may be providers but the provider.response might be rejected. Check "AppointmentProviderResponse".

api/src/main/java/org/openmrs/module/appointments/service/impl/AppointmentsServiceImpl.java
```diff
@@ -140,6 +168,10 @@ public Appointment getAppointmentByUuid(String uuid) {
 
     @Override
     public void changeStatus(Appointment appointment, String status, Date onDate) throws APIException{
+        if (!validateIfUserHasSelfOrAllAppointmentsAccess(appointment)) {
```
angshu: is this enough? For example, a provider may be one of the providers, but he can not "MANAGE APPOINTMENT" and/or this appointment was not created by him. - in such case, is he allowed to change Status?

api/src/main/java/org/openmrs/module/appointments/service/impl/AppointmentsServiceImpl.java
```diff
@@ -163,6 +195,10 @@ public void changeStatus(Appointment appointment, String status, Date onDate) th
 
     @Override
     public void undoStatusChange(Appointment appointment) throws APIException{
+        if (!validateIfUserHasSelfOrAllAppointmentsAccess(appointment)) {
```
angshu: Same question as above.

api/src/main/java/org/openmrs/module/appointments/service/impl/AppointmentsServiceImpl.java
nehagupta2123: We are not considering the creator of the appointment as of now as that would involve more changes in the product. If the provider is one of the providers of the appointment, he/she should be able to change the status of the appointment. Lets say, if there are multiple providers for an appointment and patient was attended by only one of those providers, the provider should be able to change the status of the appointment in such a case else there has to be some notification system to do the same.

api/src/main/java/org/openmrs/module/appointments/service/impl/AppointmentsServiceImpl.java
nehagupta2123: will review the code once.

api/src/main/java/org/openmrs/module/appointments/service/impl/AppointmentsServiceImpl.java
angshu: I am ok with that. Please mention this explicitly in the JIRA card. Saying out of scope for no - if a provider is one of the assignees in the appointment, then he can change status.

api/src/main/java/org/openmrs/module/appointments/service/impl/AppointmentsServiceImpl.java
```diff
@@ -52,8 +60,30 @@ public void setAppointmentAuditDao(AppointmentAuditDao appointmentAuditDao) {
...
     }
 
+    private boolean validateIfUserHasSelfOrAllAppointmentsAccess(Appointment appointment) {
+        return Context.hasPrivilege(MANAGE_APPOINTMENTS) ||
+                isAppointmentNotAssignedToAnyProvider(appointment) ||
+                isCurrentUserSamePersonAsOneOfTheAppointmentProviders(appointment.getProviders());
+    }
+
+    private boolean isAppointmentNotAssignedToAnyProvider(Appointment appointment) {
+        return isNull(appointment.getProviders()) ||
```
angshu: All these can be pushed to the appointment.getProvidersWithResponse() - right? which will make the "isAppointmentNotAssignedToAnyProvider()" redundant?

PR #10, opened 2019-04-17
API to get appointments for a date range which is used for Appointments Weekly View 

sowmika148:
[PRODUCT JIRA CARD LINK](https://bahmni.atlassian.net/browse/BAH-663)

### DIFF
--- a/api/src/main/java/org/openmrs/module/appointments/dao/impl/AppointmentDaoImpl.java
+++ b/api/src/main/java/org/openmrs/module/appointments/dao/impl/AppointmentDaoImpl.java
@@ -8,6 +8,7 @@ import org.hibernate.sql.JoinType;
 import org.openmrs.module.appointments.dao.AppointmentDao;
 import org.openmrs.module.appointments.model.Appointment;
 import org.openmrs.module.appointments.model.AppointmentServiceDefinition;
+import org.openmrs.module.appointments.model.AppointmentSearch;
 import org.openmrs.module.appointments.model.AppointmentServiceType;
 import org.openmrs.module.appointments.model.AppointmentStatus;
 import org.springframework.transaction.annotation.Transactional;
@@ -115,6 +116,15 @@ public class AppointmentDaoImpl implements AppointmentDao {
         if (endDate != null) {
             criteria.add(Restrictions.lt("endDateTime", endDate));
         }
+        return criteria.list();
+    }
+
+    @Override
+    public List<Appointment> search(AppointmentSearch appointmentSearch) {
+        Criteria criteria = sessionFactory.getCurrentSession().createCriteria(Appointment.class);
+        criteria.add(Restrictions.eq("voided", false));
+        Date maxEndDate = new Date(appointmentSearch.getEndDate().getTime() + TimeUnit.DAYS.toMillis(1));
+        criteria.add(Restrictions.between("startDateTime", appointmentSearch.getStartDate(), maxEndDate));
         return criteria.list();
     }
 }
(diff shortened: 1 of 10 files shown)


api/src/main/java/org/openmrs/module/appointments/dao/impl/AppointmentDaoImpl.java
```diff
@@ -117,4 +118,13 @@ public Appointment getAppointmentByUuid(String uuid) {
         }
         return criteria.list();
     }
+
+    @Override
+    public List<Appointment> search(AppointmentSearch appointmentSearch) {
+        Criteria criteria = sessionFactory.getCurrentSession().createCriteria(Appointment.class);
+        criteria.add(Restrictions.eq("voided", false));
+        Date maxEndDate = new Date(appointmentSearch.getEndDate().getTime());
```
angshu: - Any particular reason, for endDate - we call - appointmentSearch.getEndDate().getTime(), 
While for startDate its just appointmentSearch.getStartDate().

api/src/main/java/org/openmrs/module/appointments/dao/impl/AppointmentDaoImpl.java
```diff
@@ -117,4 +118,13 @@ public Appointment getAppointmentByUuid(String uuid) {
         }
         return criteria.list();
     }
+
+    @Override
+    public List<Appointment> search(AppointmentSearch appointmentSearch) {
```
angshu: The function seems to suggest that it takes a generic AppointmentSearch object, which I am assuming is created for future extensibility purpose. However, subsequently in the code, we seem to assume that the start and end date are always given. 
- Are these start and end dates mandatory for search? If not, then we should add criteria conditionally.

omod/src/main/java/org/openmrs/module/appointments/web/controller/AppointmentsController.java
```diff
@@ -70,4 +84,13 @@
         }
     }
 
+    @RequestMapping(method = RequestMethod.POST, value = "search")
+    @ResponseBody
+    public List<AppointmentDefaultResponse> search(@Valid @RequestBody AppointmentSearch appointmentSearch) {
+        List<Appointment> appointments = appointmentsService.search(appointmentSearch);
+        if(isNull(appointments)){
```
angshu: - How can we assume that if the result is null, then start or end date is not provided??
- If start and endDate are mandatory for the POST request, then we should have validations done of the appointmentSearch object before calling the service method. 
- why annotate with @valid?  if you want to employ validation by spring, then you should appropriately annotate the model  attributes (e.g. @NotNull etc). Spring will apply the validation as per the attribute annotations and you will get appropriate validated response, without the controller code even called for execution.

api/src/main/java/org/openmrs/module/appointments/dao/impl/AppointmentDaoImpl.java
sowmika148: In the previous commit, we added a millisecond to it. So converted to time and added a millisecond. Now we are doing that at another place but forgot to remove that declaration. Can be removed now

omod/src/main/java/org/openmrs/module/appointments/web/controller/AppointmentsController.java
bsneha90: Currently all the APIs are using @Valid, so can we go ahead and merge this? And have a card in Backlog to refactor this throughout appointment scheduling?

angshu: @valid is useless, unless you are annotating the request binding object. That should be fixed. This should not take much time. (I think I have spoken to Sowmika and told her what to do as well). Its simple spring Validator interface implementation, if the JSR validations don't kick in.

angshu: Doing this now, when the PR contributor has most context, and when the testing would be done would be the best.

PR #11, opened 2019-06-20
Sowmika, Vineela, Sneha | BAH-846 | Multiple provider fix

bsneha90:
Fix to following issue due to multiple provider:

Issue 1 - Duplicate Appointments:

1. Create an appointment with a provider.2. Now, click on edit of that appointment and remove the provider and add a new provider and save.3. It will retain the old appointment on the calendar and also show a new appointment. The list view works correctly.

Issue 2 - Cannot edit provider name again:

1. Create an appointment with a provider.2. Now, click on edit of that appointment and remove the provider and add a new provider and save.

3. Now, click on edit of that appointment and remove the provider and add a new provider, we will not be able to add it.

4. If we click on save without a provider name, it will not edit the current chosen appointment, but the duplicate appointment created in step 2. And we can still see provider in list view.


binduak:
Closing this PR as the fix for the issue already there in the bahmniapps.

PR #16, opened 2019-11-04
AP-20 | Recurring appointment API changes

bsneha90:
Find the details here - https://bahmni.atlassian.net/browse/AP-20


api/src/main/java/org/openmrs/module/appointments/helper/AppointmentServiceHelper.java
```diff
@@ -0,0 +1,100 @@
...
+import org.openmrs.module.appointments.validator.AppointmentValidator;
+import org.springframework.stereotype.Component;
+
+import java.io.IOException;
+import java.util.ArrayList;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+@Component
```
mddubey: Is this really a component? To me, it looks like a simple helper class providing few static methods.
Are we mocking `AppointmentServiceHelper` in Unit tests?

PR #24, opened 2020-01-07
AP-37 | Sukreet | Add migration to create 'Appointments: Invite Providers' privilege.

sukreet:


api/src/main/resources/liquibase.xml
```diff
@@ -548,5 +548,17 @@
             <column name="initial_appointment_status" type="varchar(45)"/>
         </addColumn>
     </changeSet>
-
+    <changeSet id="202001021618" author="Sukreet">
+        <preConditions onFail="MARK_RAN">
+            <sqlCheck expectedResult="0">
+                SELECT COUNT(*) FROM privilege WHERE privilege = 'Appointments: Add Providers';
```
bsneha90: Will 'Appointments: Invite Providers' make more sense?

sukreet: @bsneha90  Sure, will update this PR

sukreet: Done

PR #33, opened 2020-05-06
Fix nexus validations;Upgrade version to 1.3.0-SNAPSHOT

mddubey:
This PR
- [x] Adds Project URL, Licenses, Developer and Github repo information.
- [x] Adds Maven Source Plugin
- [x] Adds Maven Javadoc Plugin
- [x] Adds Maven gpg Plugin
- [x] Update version to 1.3.0-SNAPSHOT.

### DIFF
--- a/api/src/main/java/org/openmrs/module/appointments/AppointmentsActivator.java
+++ b/api/src/main/java/org/openmrs/module/appointments/AppointmentsActivator.java
@@ -25,16 +25,10 @@ public class AppointmentsActivator extends BaseModuleActivator {
 	
 	private Log log = LogFactory.getLog(this.getClass());
 	
-	/**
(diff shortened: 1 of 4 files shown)


api/src/main/java/org/openmrs/module/appointments/AppointmentsActivator.java
```diff
@@ -25,16 +25,10 @@
 	
 	private Log log = LogFactory.getLog(this.getClass());
 	
-	/**
-	 * @see org.openmrs.module.Activator#startup()
```
mddubey, line 29: We had to remove these lines because there were errors from `mvn-javadoc-plugin` as `Cannot find definition`.

PR #44, opened 2020-11-27
BAH-1093 : Tele-consultation feature merge

praveenadayanand:


api/src/main/resources/liquibase.xml
```diff
@@ -195,6 +195,16 @@
         </sql>
     </changeSet>
 
+    <changeSet id="create-column-teleconsultation-202008241355" author="Vishal">
```
angshu, line 198: precondition should be whether the column exists.

api/src/main/resources/liquibase.xml
```diff
@@ -195,6 +195,16 @@
...
+    <preConditions onFail="MARK_RAN">
+            <tableExists tableName="patient_appointment" />
+    </preConditions>
+    <sql>
```
angshu, line 202: use liquibase "addColumn" tag instead of SQL

api/src/main/resources/liquibase.xml
```diff
@@ -342,6 +352,14 @@
         <sqlFile path="patientUpcomingAppointments.sql"/>
     </changeSet>
 
+    <changeSet id="global-property-upcoming-appointments-sql-update" author="Shankar, Mahesh">
```
angshu, line 355: add global property in config.xml to add global property. 
example: https://github.com/Bahmni/openmrs-module-appointments/blob/master/omod/src/main/resources/config.xml#L61

api/src/main/resources/liquibase.xml
```diff
@@ -576,4 +594,10 @@
             <column name="description" value="URL pattern to use for published Recurring appointment events. Default is /openmrs/ws/rest/v1/recurring-appointments?uuid={uuid}. If you change default value, please add your custom implementation for the given URL"/>
         </insert>
     </changeSet>
+    <changeSet id="default-202009011116" author="Deedee Lee">
```
angshu, line 597: add precondition to check whether email as person attribute type exists

api/src/main/resources/patientUpcomingAppointmentsUpdate.sql
```diff
@@ -0,0 +1,23 @@
+UPDATE global_property
```
angshu, line 1: rename to "patientUpcomingAppointments_v2.sql"

angshu (changes requested):

api/src/main/java/org/openmrs/module/appointments/service/impl/AppointmentsServiceImpl.java
```diff
@@ -10,40 +10,27 @@
...
-import java.util.Optional;
-import java.util.Set;
+import java.util.*;
 import java.util.stream.Collectors;
 
 import static java.util.Objects.isNull;
-import static org.openmrs.module.appointments.constants.PrivilegeConstants.MANAGE_APPOINTMENTS;
-import static org.openmrs.module.appointments.constants.PrivilegeConstants.MANAGE_OWN_APPOINTMENTS;
-import static org.openmrs.module.appointments.constants.PrivilegeConstants.RESET_APPOINTMENT_STATUS;
+import static org.openmrs.module.appointments.constants.PrivilegeConstants.*;
```
angshu, line 29: replace with single line imports

angshu (changes requested):

api/src/main/resources/liquibase.xml
swatigogia2020: This is adding value in the table name global_property and not a global property used in java code.

api/src/main/java/org/openmrs/module/appointments/service/impl/AppointmentsServiceImpl.java
praveenadayanand: @angshu We will incorporate this as part of our release 2.

api/src/main/resources/liquibase.xml
praveenadayanand: Done

api/src/main/resources/liquibase.xml
praveenadayanand: Done

angshu (approved):

PR #48, opened 2020-12-21
Change defaultValue constraint in liquibase.xml to defaultValueBoolean 

praveenadayanand:


angshu:
you can not change the content of a changeset without changing the id, right? if run on already run database, this will fail? This is taken care of. can you please close the PR.

praveenadayanand:
> you can not change the content of a changeset without changing the id, right? if run on already run database, this will fail? This is taken care of. can you please close the PR.

Yes Angshu, we verified that the changeset did not run on product-qa04. The liquibaselock table does not have this changeset and so we feel that the changeset would work without modifying the id. On the other hand, in our QA environment, since the changeset already was applied, we created a new changeset and tested it out.

PR #51, opened 2021-01-21
BAH-1133: Setting up time zone sensitive tests, adjusting GitHub Actions.

mks-d:
#### Ticket
https://bahmni.atlassian.net/browse/BAH-1133

### DIFF
--- a/api/src/test/java/org/openmrs/module/appointments/util/DateUtilTest.java
+++ b/api/src/test/java/org/openmrs/module/appointments/util/DateUtilTest.java
@@ -1,6 +1,11 @@
 package org.openmrs.module.appointments.util;
 
-import org.junit.Test;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotNull;
+import static org.junit.Assert.assertNull;
+import static org.junit.Assert.assertTrue;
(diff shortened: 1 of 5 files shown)


api/src/test/java/org/openmrs/module/appointments/util/DateUtilTest.java
```diff
@@ -85,6 +85,7 @@ public void shouldReturnEndOfDay() {
 
     @Test
     public void shouldConvertDateToMilliSeconds() throws ParseException {
+    	TimeZone.setDefault(TimeZone.getTimeZone("IST"));
```
binduak, line 88: @mks-d we can have this timezone configuration as part of the mvn build command ?

mks-d: @binduak it is not ok to have Maven builds that are time zone dependent. The whole point of this ticket is to get rid of that situation.

Devs, when writing test cases that are time zone dependent, should be aware of it and should explicitly make this part of their test setups.

binduak: my bad...i missed going through the card before having a look at PR. Totally makes sense

binduak (approved):

PR #52, opened 2021-01-28
Bindu | BAH-1136 | Change query to get provider from patient appointment provider table

binduak:
https://bahmni.atlassian.net/browse/BAH-1136


buvaneswari-arun (approved):
Approving

PR #53, opened 2021-02-04
Bindu | BAH-1136 | change the changeset id to fix liquibase migration failure issue

binduak:
https://bahmni.atlassian.net/browse/BAH-1136

### DIFF
--- a/api/src/main/resources/liquibase.xml
+++ b/api/src/main/resources/liquibase.xml
@@ -347,7 +347,7 @@
         <sqlFile path="patientPastAppointments.sql"/>
     </changeSet>
 
-    <changeSet id="global-property-upcoming-appointments-sql" author="Maharjun, Shireesha">
+    <changeSet id="global-property-upcoming-appointments-sql-04022021" author="Maharjun, Shireesha">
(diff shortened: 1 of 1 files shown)


praveenadayanand (changes requested):
@binduak In the V2 version of the upcoming-appointments.sql file we have included one more field(teleconsultation) to be fetched from the DB.

binduak:
@praveenadayanand then why do we require V2 version of sql file. We can have teleconsultation changes in the same file ?

praveenadayanand:
> @praveenadayanand then why do we require V2 version of sql file. We can have teleconsultation changes in the same file ?

@binduak Yes you are correct, the teleconsultation field in V2 file can be included in V1 sql itself.

angshu:
1 issue on this. this SQL was there in Release 1.1 and also for 1.2. so changing the changeset Id and with the precondition, existing SQLs will never get updated. Ideally, this should be ideally written as update statement (if property exists), and if does not exist, then do an insert. For now, i am approving this and merging this, to unblock the QA env. however, please raise another PR to fix this.

angshu (approved):

PR #55, opened 2021-02-05
BAH-1140|[Mahesh] Handling of missing email configuration

praveenadayanand:


api/src/main/resources/liquibase.xml
```diff
@@ -364,6 +364,14 @@
         <sqlFile path="patientUpcomingAppointments_v2.sql"/>
     </changeSet>
 
+    <changeSet id="global-property-upcoming-appointments_updated" author="Gogia, Swati">
+        <preConditions onFail="MARK_RAN">
+            <sqlCheck expectedResult="1">select count(*) from global_property where property='bahmni.sqlGet.upComingAppointments'</sqlCheck>
```
binduak, line 369: @praveenadayanand  should the version be v2 or v3 ? also can you remove the change sets for the previous sql from liquibase.xml file. Lets keep the old file as it is. But lets remove the respective change set from the liquibase.xml file.

praveenadayanand: @binduak After PR #50 is merged this PR would need to be updated to pick up those changes. Correct me if I am wrong here @swatigogia2020

buvaneswari-arun (approved):

PR #61, opened 2021-03-23
AP-77: Changed characters escaping in Liquibase SQL files.

icrc-toliveira:
Issue: https://bahmni.atlassian.net/browse/AP-77

A MYSQL error was found when using openMRS 2.4.0.
The cause seems to be related with the functions inside the files:
* patientPastAppointments.sql
* patientPastAppointments_v2.sql
* patientUpcomingAppointments.sql
* patientUpcomingAppointments_v2.sql

After a closer look, root of the error seems to be the way some characters are escaped.

Solution:
The solution seems too be using different forms of "" and '' instead of relying on \ to do the character escaping


icrc-toliveira:
"A MYSQL error was found when using openMRS 2.4.0."
This problem seems to result into the following error at runtime:

image

dkayiwa:
@icrc-toliveira i see that you proposed a solution in the pull request description. Did you try it out?

icrc-toliveira:
@dkayiwa I did update the PR
Instead of duplicating the files, I used the old ones and placed the tag `runOnChange="true"` in the changeSet definition
This way, even if the file was already executed by the liquidbase, it still will run.

dkayiwa:
@icrc-toliveira do you have any changes that require to run again for servers where this changeset was run before?

icrc-toliveira:
@dkayiwa 
As a response to "do you have any changes that require to run again for servers where this changeset was run before?"
I don´t think so, since the problem happens during the insertion (If query is already present there shouldn´t be a problem). 
However, if the new version has a different checksum it will throw an error, that why I used "runOnChange='true'".

dkayiwa:
Does the checksum also include the contents of the referenced file?

icrc-toliveira:
@dkayiwa 
Not  sure if I understand the question
However from what I understand, the liquidbase checksum is generated based on the content of the sql file.

dkayiwa:
@icrc-toliveira are you aware of these merge conflicts?

icrc-toliveira:
@dkayiwa The conflicts should now be solved

dkayiwa:
@icrc-toliveira looks good. If i had merge privileges for this repository, i would have done the needful. 😊

api/src/main/resources/patientPastAppointments.sql
```diff
@@ -1,10 +1,9 @@
 INSERT INTO global_property (property, property_value, description, uuid)
- VALUES ('bahmni.sqlGet.pastAppointments',
-"SELECT
+VALUES ('bahmni.sqlGet.pastAppointments','SELECT
      app_service.name                                                                                AS `DASHBOARD_APPOINTMENTS_SERVICE_KEY`,
      app_service_type.name                                                                           AS `DASHBOARD_APPOINTMENTS_SERVICE_TYPE_KEY`,
-     DATE_FORMAT(start_date_time, \"%d/%m/%Y\")                                                        AS `DASHBOARD_APPOINTMENTS_DATE_KEY`,
-     CONCAT(DATE_FORMAT(start_date_time, \"%l:%i %p\"), \" - \", DATE_FORMAT(end_date_time, \"%l:%i %p\")) AS `DASHBOARD_APPOINTMENTS_SLOT_KEY`,
```
binduak, line 7: @icrc-toliveira @mks-d what is the mysql version that you are facing error with ?

icrc-toliveira: Hi @binduak 
I´m facing this issue with "Mysql  Ver 14.14 Distrib 5.7.31"

binduak (approved):

api/src/main/resources/patientPastAppointments.sql
binduak: Tested migration changes with mysql 5.6 Distrib, and the changes are working fine. Approved the PR.

PR #65, opened 2021-06-17
AP-77: Changed characters escaping in Liquibase SQL files.

icrc-toliveira:
Note:  Some files contained an error in the SQL (patientPastAppointments.sql / patientUpcomingAppointments.sql). 

Issue: https://bahmni.atlassian.net/browse/AP-77

A MYSQL error was found when using openMRS 2.4.0.
The cause seems to be related with the functions inside the files:

patientPastAppointments.sql
patientPastAppointments_v2.sql
patientUpcomingAppointments.sql
patientUpcomingAppointments_v2.sql
After a closer look, root of the error seems to be the way some characters are escaped.

Solution:
The solution seems too be using different forms of "" and '' instead of relying on \ to do the character escaping

### DIFF
--- a/api/src/main/resources/patientPastAppointments.sql
+++ b/api/src/main/resources/patientPastAppointments.sql
@@ -1,10 +1,9 @@
 INSERT INTO global_property (property, property_value, description, uuid)
- VALUES ('bahmni.sqlGet.pastAppointments',
-"SELECT
+VALUES ('bahmni.sqlGet.pastAppointments','SELECT
      app_service.name                                                                                AS `DASHBOARD_APPOINTMENTS_SERVICE_KEY`,
      app_service_type.name                                                                           AS `DASHBOARD_APPOINTMENTS_SERVICE_TYPE_KEY`,
-     DATE_FORMAT(start_date_time, \"%d/%m/%Y\")                                                        AS `DASHBOARD_APPOINTMENTS_DATE_KEY`,
-     CONCAT(DATE_FORMAT(start_date_time, \"%l:%i %p\"), \" - \", DATE_FORMAT(end_date_time, \"%l:%i %p\")) AS `DASHBOARD_APPOINTMENTS_SLOT_KEY`,
+     DATE_FORMAT(start_date_time, "%d/%m/%Y")                                                        AS `DASHBOARD_APPOINTMENTS_DATE_KEY`,
+     CONCAT(DATE_FORMAT(start_date_time, "%l:%i %p"), " - ", DATE_FORMAT(end_date_time, "%l:%i %p")) AS `DASHBOARD_APPOINTMENTS_SLOT_KEY`,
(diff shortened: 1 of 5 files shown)


api/src/main/resources/patientPastAppointments.sql
```diff
@@ -1,22 +1,22 @@
...
+     DATE_FORMAT(start_date_time, "%d/%m/%Y")                                                        AS `DASHBOARD_APPOINTMENTS_DATE_KEY`,
+     CONCAT(DATE_FORMAT(start_date_time, "%l:%i %p"), " - ", DATE_FORMAT(end_date_time, "%l:%i %p")) AS `DASHBOARD_APPOINTMENTS_SLOT_KEY`,
      CONCAT(pn.given_name, ' ', pn.family_name)                                                      AS `DASHBOARD_APPOINTMENTS_PROVIDER_KEY`,
      pa.status                                                                                       AS `DASHBOARD_APPOINTMENTS_STATUS_KEY`
 FROM
    patient_appointment pa
    JOIN person p ON p.person_id = pa.patient_id AND pa.voided IS FALSE
    JOIN appointment_service app_service
      ON app_service.appointment_service_id = pa.appointment_service_id AND app_service.voided IS FALSE
+   LEFT JOIN patient_appointment_provider pap on pa.patient_appointment_id = pap.patient_appointment_id AND (pap.voided=0 OR pap.voided IS NULL)
```
icrc-toliveira, line 14: Hi @binduak 

After taking a look, it seems that the use of "pap" instead of "pa", is intended.
The problem was that the files "patientPastAppointments.sql" and "patientUpcomingAppointments.sql" were both missing a join. 
Keep in mind that this join used to be present in past versions, however was removed a couple PR's ago.

Also another reason that lead me to believe the use of "pap" was intended, was its presence on "patientPastAppointments_v2.sql" and "patientUpcomingAppointments_v2". Files us purpose is to update the queries present in "patientPastAppointments.sql" and "patientUpcomingAppointments.sql"

icrc-toliveira: @binduak

Side note, I installed the module using the OpenMrs SDK with MySQL 5.7 and everything seems to work after the fix mentioned above. 
However if you find any issue pls let me know

binduak: Thanks @icrc-toliveira the reason we don't have join in patientPastAppointments.sql file for pap is, we didn't have patient_appointment_provider table as part of bahmni 0.92 or before.
So we don't require pap in this file at all. 
Will remove the pap part once this PR gets merged.

binduak: @mks-d @icrc-toliveira @angshu I have tested sql changes with mysql 5.6 and 5.7 this time and everything seems working fine. I will go ahead and merge the PR !

binduak (approved):

icrc-toliveira:
Hi @binduak, @mks-d 

If there is no problem could we merge this PBI?

PR #71, opened 2021-12-03
BAH-1209: Fetch appointments for non-voided patients only (Fix Regression)

icrc-toliveira:
Issue: https://bahmni.atlassian.net/browse/AP-78

Appointments module seems not to be handling voided patients correctly. 

Regression: 
While in the past validation for Patient = Voided was enough 
Now, it seems that a Person can be voided while the Patient is still not voided. 
This causes an error.

Fix suggested:
Also validate for Person = voided


ibacher:
@angshu Can you take a look at this or assign it to the appropriate person to review?

PR #72, opened 2021-12-23
BAH-1280: Sms implementation

shrey-beehyv:


api/pom.xml
```diff
@@ -149,6 +149,13 @@
 			<scope>provided</scope>
 		</dependency>
 		<!-- End OpenMRS core -->
+
+		<!-- https://mvnrepository.com/artifact/com.squareup.okhttp/okhttp -->
+		<dependency>
```
angshu, line 154: Do we need another http client? We already use apache's httpclient

api/src/main/java/org/openmrs/module/appointments/notification/impl/DefaultSmsSender.java
```diff
@@ -0,0 +1,24 @@
+package org.openmrs.module.appointments.notification.impl;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.openmrs.api.AdministrationService;
+import org.openmrs.module.appointments.notification.SmsSender;
+
+import java.util.List;
+
+public class DefaultSmsSender implements SmsSender {
```
angshu, line 10: This one does not do anything. Any reason for it to be here? just the interfac "SmsSender" should not be enough? We can always check if the dependency is provided in code and not do anything.

PR #73, opened 2022-01-19
AP-82 - Timezone support for service definitions start and end times

icrc-psousa:
**Ticket:** [AP-82](https://bahmni.atlassian.net/browse/AP-82)
**Description:** Reviewed services saving / loading to cope with timezones
- Include Date and timezone info when converting Time to String so that, when it's consumed, it can be properly converted to other timezones. We'll need to add current date to the time and not something like "January 01, 1970" to have the right hour whether we are on daylight saving time or not.


ibacher:
@angshu @binduak Any chance this can be reviewed?

icrc-psousa:
@angshu @binduak PR was adapted to have DB times saved with server time instead of UTC as discussed. Would it be possible to review this?

PR #74, opened 2022-01-27
AP-12: Removed hard dependency on Atom feed.

ibacher:
This refactors things so that the module does not require AtomFeed. This is done by:

1. Marking the module as "aware of" AtomFeed rather than requiring AtomFeed
2. Moving the Advice classes which need AtomFeed class into their own separate Jar
3. Ensuring that this Jar is only loaded when the AtomFeed module is present

This should ensure that the module continues to work in Bahmni but can be used in non-Bahmni versions of OpenMRS that would prefer not to use AtomFeed.


gsluthra:
@binduak - Does this affect your team too? Can your team also provide review comments to this?

angshu (approved):

binduak:
Hi @angshu @gsluthra The data from openmrs to Bahmni-mart sync works based on the atomfeed events. So we need atomfeed events for appointments module. But from card we could see that the events may/may not generate with the PR changes.

gsluthra:
@binduak From what I understand, it should work for Bahmni with atomfeed+advice jars being present. You can test and validate. For others who need appointment module outside of Bahmni, they can choose to not have these jars (and this module should work for them). Yes-in that case Bahmni mart won't be available to them, which should be fine.

binduak:
@ibacher @gsluthra @angshu atomfeed events are generating with the PR changes for appointments. Approved the PR !

binduak (approved):

PR #86, opened 2022-09-05
AP-87: Enforcing BOOLEAN datatype on boolean fields

Ruhanga:
Ticket: https://bahmni.atlassian.net/browse/AP-87


gsluthra:
Modifying an existing liquibase changeset is not recommended, since Databases which have already run these changesets, will ignore this change. We may need to consider a schema modification, or atleast rename the changeset to act as new. Or, is the intention to ignore these changesets in existing systems? 

Basically -- need to discuss
a) What happens when the new omod is deployed in an existing system which already has these tables as Boolean.
b) What happens when the omod is deployed in a new system (i guess this will work fine, although changeset name points to old changeset timeperiod)

binduak:
@Ruhanga can you share the stack trace for the "runtime unsaved transient entity error" mentioned in the card

Ruhanga:
Thanks for your time reviewing this issue @gsluthra, @binduak. 
>Basically -- need to discuss
a) What happens when the new omod is deployed in an existing system which already has these tables as Boolean.

Following your comment above @gsluthra, it's much more precise adding changesets to modify the existing schemas respectively for consistency, and this is the whole importance of using liquibase indeed. Otherwise the changed changeset would fail for newer version of the omod.

Ruhanga (commented):
@binduak, here is [the error](https://pastebin.com/raw/ULey6YbQ) which I think is resolve by [this commit](https://github.com/Bahmni/openmrs-module-appointments/commit/bbe2d0676ff62ea2f80944516fd0bf038e2e782a) in which case I'll revert the changes on the hibernate mappings and update the ticket description too.

Ruhanga:
@gsluthra, @binduak, could you review again? Thank you.

binduak (approved):

PR #89, opened 2022-09-21
BAH-2180|sms-alerts-appointments-booking

shilpa-iplit:
Sample contents for sms-config.properties file to be placed in openmrs data directory

sms.key.username=username
sms.key.password=password
sms.key.phonenumber=destination
sms.key.message=message
sms.key.sender=source
sms.key.entity.identifier=entityid
sms.key.template.identifier=tempid

sms.phonenumber.length=10

sms.url=http://103.16.101.52:8080/bulksms/bulksms?
sms.username=ints-wemdco
sms.password=IPLit607
sms.sender=IPLits
sms.entity.identifier=1201159240485243492
sms.params.addons=type=0&dlr=1

sms.appointment.walkin.template.identifier=1207161751603402657
sms.appointment.teleconsultation.template.identifier=1207162489034032086

sms.appointment.walkin.message.template=Dear %s,\n Your appointment is booked for %s on %s at %s by Bahmni Hospital (powered by IPLit)
sms.appointment.teleconsultation.message.template=Dear %s,\nTele-Consultation booked for %s %s\nVideo link: %s Bahmni Hospital(IPLit)

### DIFF
--- a/api/src/main/java/org/openmrs/module/appointments/notification/SmsSender.java
+++ b/api/src/main/java/org/openmrs/module/appointments/notification/SmsSender.java
@@ -1,0 +1,7 @@
+package org.openmrs.module.appointments.notification;
+
+import java.util.Date;
+
+public interface SmsSender {
+    void send(String recipientMobileNumber, String patientName, String smsType, String serviceName, Date startDateTime, String teleLink);
+}
--- a/api/src/main/java/org/openmrs/module/appointments/notification/impl/DefaultSmsSender.java
+++ b/api/src/main/java/org/openmrs/module/appointments/notification/impl/DefaultSmsSender.java
@@ -1,0 +1,180 @@
+package org.openmrs.module.appointments.notification.impl;
+
+import org.apache.commons.lang.StringUtils;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.openmrs.api.AdministrationService;
+import org.openmrs.module.appointments.notification.SmsSender;
+import org.openmrs.util.OpenmrsUtil;
+
+import java.io.IOException;
+import java.net.HttpURLConnection;
+import java.net.URL;
+import java.net.URLEncoder;
+import java.nio.file.Files;
+import java.nio.file.Path;
+import java.nio.file.Paths;
+import java.text.SimpleDateFormat;
+import java.util.Date;
+import java.util.Properties;
+
+public class DefaultSmsSender implements SmsSender {
+
+    private Log log = LogFactory.getLog(this.getClass());
+
+    private static final String SMS_PROPERTIES_FILENAME = "sms-config.properties";
+    private static Properties smsSessionProperties = null;
+
+    SimpleDateFormat fmtTime = new SimpleDateFormat("HH:mm a");
+    SimpleDateFormat fmtDateTele = new SimpleDateFormat("dd/MM/yyyy");
+
+    private AdministrationService administrationService;
+
+    public DefaultSmsSender(AdministrationService administrationService) {
+        this.administrationService = administrationService;
+    }
+
+    @Override
+    public void send(String recipientMobileNumber, String patientName, String smsType, String serviceName, Date startDateTime, String teleLink) {
+        HttpURLConnection uc = null;
+        try {
+        	getSession();
+
+        	String password = "";
+            String addOnParams = "";
+            String entityId = "";
+            String templateId = "";
+            String msgText = "";
+
+            recipientMobileNumber = recipientMobileNumber.replace("+", "");
+            int recipientPhoneLengthRequired = Integer.parseInt(smsSessionProperties.getProperty("sms.phonenumber.length"));
+            if (recipientMobileNumber.length()>recipientPhoneLengthRequired) {
+            	int indexStart = recipientMobileNumber.length() - recipientPhoneLengthRequired;
+            	recipientMobileNumber = recipientMobileNumber.substring(indexStart);
+            }
+            String dateSuffixText = getDateWithSuffix(startDateTime.getDate());
+            String dateApptPattern = String.format("MMMM '%s', yyyy", dateSuffixText);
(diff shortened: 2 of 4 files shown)


api/src/main/java/org/openmrs/module/appointments/notification/SmsSender.java
```diff
@@ -0,0 +1,7 @@
+package org.openmrs.module.appointments.notification;
+
+import java.util.Date;
+
+public interface SmsSender {
+    void send(String recipientMobileNumber, String patientName, String smsType, String serviceName, Date startDateTime, String teleLink);
```
gsluthra, line 6: I was hoping the interface is Channel vocabulary based and not Appointment vocabulary based. Something like:

MessageSender.send(String phoneNumber, String fullMessage).

Then SMS Sender, Whatsapp Sender, etc can easily implement this interface. They don't care if the message is goign to patient, doctor, customer, relative, etc. They just send a message over the appropriate channel to the number. Its the caller of these classes that constructs the message and delegates the "sending" to the channel implementation.

So.. Appointment ---delegates to ---> AppointmentMessage class ---delegates to--> SMSSender class ---> sends SMS

Similarly,
Appointment ---delegates to ---> AppointmentMessage class ---delegates to (based on config)--> WhatsappSender class ---> sends SMS

Then we write unit tests for AppointmentMessage class which checks if the construction of message is correct based on passed parameters like PatientName, Service Name, Date, Link.

With this logic, the MessageSender interface and its implementations are reusable in the context of different messages (Appointments, Registration, Teleconsult, Doctor-Doctor consult, etc). 

For this JIRA ticket, we would only implement SMS (and not whatzapp classes).

api/src/main/java/org/openmrs/module/appointments/notification/impl/DefaultSmsSender.java
```diff
@@ -0,0 +1,180 @@
...
+            int recipientPhoneLengthRequired = Integer.parseInt(smsSessionProperties.getProperty("sms.phonenumber.length"));
+            if (recipientMobileNumber.length()>recipientPhoneLengthRequired) {
+            	int indexStart = recipientMobileNumber.length() - recipientPhoneLengthRequired;
+            	recipientMobileNumber = recipientMobileNumber.substring(indexStart);
+            }
+            String dateSuffixText = getDateWithSuffix(startDateTime.getDate());
+            String dateApptPattern = String.format("MMMM '%s', yyyy", dateSuffixText);
+            SimpleDateFormat fmtDateAppt = new SimpleDateFormat(dateApptPattern);
+
+            String requestUrl = smsSessionProperties.getProperty("sms.url");
```
gsluthra, line 59: A lot of logic for string concatenation. Needs tests. See the previous comment on how this could possibly be restructured into an AppointmentMessage class, which constructs the message string.

binduak: +1

api/src/main/java/org/openmrs/module/appointments/notification/SmsSender.java
shilpa-iplit: Building of the message text requires sms template identifier. The message template format is available in SMSSender class (DefaultSmsSender) which is needed as outgoing SMS has to be conforming to this format.
Either message template is available in config file or in global properties in openmrs.
Please advise

api/src/main/java/org/openmrs/module/appointments/notification/impl/DefaultSmsSender.java
```diff
@@ -0,0 +1,180 @@
...
+                    URLEncoder.encode(smsSessionProperties.getProperty("sms.appointment.walkin.template.identifier"), "UTF-8");
+                // Dear %s,\n Your appointment is booked for %s on %s at %s by Bahmni Hospital (powered by IPLit)
+                msgText = String.format(smsSessionProperties.getProperty("sms.appointment.walkin.message.template"), patientName,
+                    serviceName, fmtDateAppt.format(startDateTime), fmtTime.format(startDateTime));
+            }
+            if (StringUtils.isNotBlank(smsSessionProperties.getProperty("sms.params.addons"))) {
+                addOnParams = "&" + smsSessionProperties.getProperty("sms.params.addons");
+            }
+            String message = "&" + smsSessionProperties.getProperty("sms.key.message") + "=" + URLEncoder.encode(msgText, "UTF-8");
+            requestUrl += apiKey + message + sender + number + password + entityId + templateId + addOnParams;
```
angshu, line 88: sending the msg itself through the URL  is not a standard way of passing the sms. Can you check if you can pass through the body/payload .. 
Also, its probably better to use a closeable HTTP Client, and you can use try with resources block

api/src/main/java/org/openmrs/module/appointments/notification/impl/DefaultSmsSender.java
```diff
@@ -0,0 +1,180 @@
...
+
+            URL url = new URL(requestUrl);
+            uc = (HttpURLConnection)url.openConnection();
+            String smsResponse = uc.getResponseMessage();
```
angshu, line 92: if you aren't logging the response, why assign to a variable?

api/src/main/java/org/openmrs/module/appointments/notification/impl/DefaultSmsSender.java
```diff
@@ -0,0 +1,180 @@
...
+            } catch (IOException e) {
+                log.error("Could not load sms properties from: " + propertyFilePath, e);
+            }
+        } else {
+            log.warn("No sms configuration defined at " + propertyFilePath);
+        }
+        return null;
+    }
+
+    private String getDateWithSuffix(int date) {
```
angshu, line 160: do we really need to hardcode st, nd etc .. why not just convert to string representation of date .. like 30 June 2022. its also not using the locale and using just english. You can leverage a date formatter

shilpa-iplit: sure. Will incorporate all the comments.

PR #91, opened 2022-10-04
Ap 84 - Add "age", "gender" and "patient identifiers" to Patient map 

icrc-toliveira:
Issue: https://bahmni.atlassian.net/browse/AP-84

Note: 
In a prior commit, the fields were not added to the listView. 
This would break the report and show fields as empty 
(As a solution we added this fields to the template)

That being said, one other issue that stem from this, is the need to show a specific auxiliary Identifier. 
This is a problem when we have multi auxiliary Identifiers. 

To keep it simple, instead of adding a bunch of  auxiliary Identifiers, we now add a auxiliaryIdentifier key, containing all the remaining auxiliary Identifiers


icrc-toliveira:
Hi @binduak 
Could you take a look to make sure you no longer find the the fields missing

Related PR: https://github.com/Bahmni/openmrs-module-appointments-frontend/pull/230

CynthiaKamau:
cc @donaldkibet

omod/src/main/java/org/openmrs/module/appointments/web/mapper/AppointmentMapper.java
```diff
@@ -252,6 +250,9 @@ private Map createPatientMap(Patient p) {
         map.put("name", p.getPersonName().getFullName());
         map.put("uuid", p.getUuid());
         map.put("identifier", p.getPatientIdentifier().getIdentifier());
+        map.put("age", p.getAge());
+        map.put("gender", p.getGender());
+        map.putAll(p.getActiveIdentifiers().stream().filter(e -> e.getIdentifierType() != null).collect(Collectors.toMap(e -> e.getIdentifierType().toString().replaceAll("[- ]", ""), e -> e.getIdentifier())));
```
angshu, line 255: What does it put in the map? 
From the code - seems the key is the "identifier type" and value is the identifier. 
Also from the replaceall expression, this seems very specific to your requirements. 

Also note, you are using "PatientIdentifierType.toString()" - which is just sending back "identifier" value - this will result in something like "HID101" : "HID101". I am not sure I understand the requirement. 

This should not be so, if you want all identifiers, then I prefer you sending as a top level attribute to the patient map object .. maybe call "identifiers" (as we are already sending the primary identifier through patient.identifier). 
If you do not require that in the map, kindly remove this line.

PR #92, opened 2022-10-11
BAH-1256 | Adding javax.mail dependency

kushboojain:
Adding dependency to solve the error of class not found exception


api/pom.xml
```diff
@@ -136,6 +136,11 @@
 			<artifactId>servlet-api</artifactId>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
```
angshu, line 139: Why do we need to include javax.mail? this module depends on openmrs core, which already bundles mail.

angshu (approved):

PR #96, opened 2022-11-09
BAH-2531 - [ Rohit ] - Resolves mail sending issue for appointments

rohit-yawalkar:
This PR is about -
 - Resolving mail sending issue for appointments
 - Jira -> https://bahmni.atlassian.net/browse/BAH-2531

### DIFF
--- a/api/src/main/java/org/openmrs/module/appointments/notification/impl/DefaultMailSender.java
+++ b/api/src/main/java/org/openmrs/module/appointments/notification/impl/DefaultMailSender.java
@@ -61,7 +61,13 @@ public class DefaultMailSender implements MailSender {
             Multipart multipart = new MimeMultipart();
             multipart.addBodyPart(mimeBodyPart);
             mail.setContent(multipart);
-            Transport.send(mail);
+
+            Transport transport = session.getTransport();
+            log.info("Sending Mail");
+            transport.connect(session.getProperty("mail.smtp.host"), session.getProperty("mail.user"), session.getProperty("mail.password"));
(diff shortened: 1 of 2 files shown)


api/src/main/java/org/openmrs/module/appointments/notification/impl/DefaultMailSender.java
```diff
@@ -88,12 +94,12 @@ private Session getSession() {
                         log.info("Could not load mail properties from application data directory. Loading from OMRS settings.");
                         sessionProperties = mailSessionPropertiesFromOMRS();
                     }
-                    final String user = sessionProperties.getProperty("mail.user");
-                    final String password = sessionProperties.getProperty("mail.password");
-                    if (StringUtils.isNotBlank(user) && StringUtils.isNotBlank(password)) {
+                    final String username = sessionProperties.getProperty("smtp.username");
```
angshu, line 97: any reason for changing the property names? the reason for such keys were to keep in sync with the OMRS global properties for mail config.

angshu: maybe the property name should be mail.smtp.user, and mail.smtp.password .. if you are changing here, make sure the same is done in the mapping in method - DefaultMailSender.mailSessionPropertiesFromOMRS()

rohit-yawalkar: @angshu, kept properties as mail.smtp.user and mail.smtp.password

PR #98, opened 2022-11-15
BAH-2531 - [ Rohit ] - Adds check to send mail only if flag is true

rohit-yawalkar:
This PR is about:
 - Adding check to send mail only if flag is true


binduak (approved):

mohan-13 (approved):
Shouldn't we consider adding a test case for the condition ?

PR #105, opened 2023-04-25
Backend migration to create dateless appointments

kavitha-sundararajan:
Database table alteration to update appointment's start and end date time to be nullable
Accept null values for start and end date times while creating appointments


riteshghiya90 (approved):
Kavitha run the migration on the data which in size is similar to prod. And it took less than a minute. Also we observed that application won't crash even if the migration took 11 minutes. Hence we are good with this migration for prod.

binduak (approved):

PR #110, opened 2023-06-01
BAH-2181 | added sms reminder for appointments

atish160384:
- Added a functionality to send SMS reminder to patients having appointment after X hours


atish160384:
This PR was raised with two templates:-
1.AppointmentReminderWithProvider
2.AppointmentReminderWithoutProvider

But  teleconsultation link was not handeled in any of them 
Later on we decider to have serpreste templates like:
1.AppointmentReminderTemplate
2.ProviderTemplate
3.TeleconsultationLinkTemplate

And appending them based upon the requirements but these changes were taken care as part of card
BAH-2962:

api/src/main/java/org/openmrs/module/appointments/service/impl/AppointmentVisitLocation.java
```diff
@@ -0,0 +1,56 @@
...
+            String baseUriProperty = AppointmentProperties.properties.getProperty("openmrs.root.uri");
+            if (baseUriProperty != null && !baseUriProperty.isEmpty()) {
+                this.openmrsBaseurl = baseUriProperty;
+                return;
+            }
+        }
+        this.openmrsBaseurl = baseURL;
+    }
+
+    public String getClinicName(String locationUuid) {
```
gsluthra, line 51: We want to use the term getFacilityName( ) in bahmni code, to avoid making it specific to clinic/hospital.. since the same code will run in all contexts.

api/src/main/java/org/openmrs/module/appointments/service/impl/AppointmentsServiceImpl.java
```diff
@@ -120,6 +122,24 @@ private boolean isCurrentUserSamePersonAsOneOfTheAppointmentProviders(Set<Appoin
...
+        Date date = appointment.getStartDateTime();
+        String service = appointment.getService().getName();
+        List<AppointmentProvider> appointmentProviderList =new ArrayList<>(appointment.getProviders());
+        List<String> providers=new ArrayList<>();
+        for (AppointmentProvider appointmentProvider: appointmentProviderList) {
+            providers.add(appointmentProvider.getProvider().getName());
+        }
+        Location location=appointment.getLocation();
+        String message = smsService.getAppointmentMessage(givenName, familyName, patientID, date, service, providers,location);
+        PersonAttribute phoneNumber = appointment.getPatient().getAttribute("phoneNumber");
```
gsluthra, line 140: Shall we check for phoneNumber first. If that is null, the rest of the code is of no use, since no message will be sent.

gsluthra: You can also consider adding a log message somewhere to say, since no mobile number found for patient, therefore didn't send SMS.

api/src/main/java/org/openmrs/module/appointments/service/impl/SMSService.java
```diff
@@ -0,0 +1,116 @@
...
+        String smsTemplate = Context.getAdministrationService().getGlobalProperty(APPOINMTMENT_REMINDER_SMS_TEMPLATE);
+        String smsTemplateNoProvider = Context.getAdministrationService().getGlobalProperty(APPOINMTMENT_REMINDER_SMS_TEMPLATE_NO_PROVIDER);
+        String smsTimeZone = Context.getMessageSourceService().getMessage(SMS_TIMEZONE, null, new Locale("en"));
+        String smsDateFormat = Context.getMessageSourceService().getMessage(SMS_DATEFORMAT, null, new Locale("en"));
+        String date =convertUTCToGivenFormat(appointmentDate, smsDateFormat, smsTimeZone);
+        String helpdeskNumber = Context.getAdministrationService().getGlobalPropertyObject("clinic.helpDeskNumber").getPropertyValue();
+        LocationTag abc = Context.getLocationService().getLocationTagByName("Visit Location");
+        List<Location> locations = Context.getLocationService().getLocationsHavingAnyTag(Collections.singletonList(abc));
+        String clinicName = (abc != null && !locations.isEmpty()) ? locations.get(0).getName() : "xxxxx";
```
gsluthra, line 55: Can we call this facilityName instead of clinicName?

api/src/main/java/org/openmrs/module/appointments/service/impl/SMSService.java
```diff
@@ -0,0 +1,116 @@
...
+        arguments.add(name + " " + familyName);
+        arguments.add(id);
+        arguments.add(date);
+        arguments.add(org.springframework.util.StringUtils.collectionToCommaDelimitedString(providers));
+        arguments.add(service);
+        arguments.add(clinicName);
+        arguments.add(helpdeskNumber);
+
+        String message;
+        if (StringUtils.isBlank(smsTemplate)) {
```
gsluthra, line 72: Consider converting this to a guard condition.. where if smsTemplate is Blank, you return and don't do anything after that. It doesn't need brackets.

api/src/main/java/org/openmrs/module/appointments/service/impl/SMSService.java
```diff
@@ -0,0 +1,116 @@
...
+                arguments.remove(3);
+                message=new MessageFormat(smsTemplateNoProvider).format(arguments.toArray()).replace("\\n", System.lineSeparator());
+            } else {
+                message =new MessageFormat(smsTemplate).format(arguments.toArray()).replace("\\n", System.lineSeparator());
+            }
+        }
+        return message;
+    }
+
+    public Object sendSMS(String phoneNumber, String message) {
```
gsluthra, line 90: Why does this method return an Object, instead of a boolean or String, or Status object?

api/src/main/java/org/openmrs/module/appointments/service/impl/SMSService.java
```diff
@@ -0,0 +1,116 @@
...
+        try {
+            SMSRequest smsRequest = new SMSRequest();
+            smsRequest.setPhoneNumber(phoneNumber);
+            smsRequest.setMessage(message);
+
+            ObjectMapper Obj = new ObjectMapper();
+            String jsonObject = Obj.writeValueAsString(smsRequest);
+            StringEntity params = new StringEntity(jsonObject);
+
+            HttpPost request = new HttpPost(Context.getMessageSourceService().getMessage(SMS_URL, null, new Locale("en")));
```
gsluthra, line 100: Locale en is hardcoded? Is this for system user?

api/src/main/resources/appointment.properties
```diff
@@ -0,0 +1,14 @@
...
+con.accept  = application/json
+con.setDoOutput = true
+con.setDoInput = true
+
+
+openmrs.host = localhost
+openmrs.port = 8080
+openmrs.root.uri=http://${openmrs.host}:${openmrs.port}/openmrs/ws/rest/v1
+openmrs.auth.uri=${openmrs.root.uri}/session
+openmrs.user= systemuser
```
gsluthra, line 11: Can we plz make this user as `system-user`.

api/src/main/resources/liquibase.xml
```diff
@@ -289,6 +289,22 @@
         </sql>
     </changeSet>
 
+    <changeSet id="create-appointment-status-change-task-201709121099999" author="Anubhav">
```
gsluthra, line 292: Author should be `Bahmni`. We stopped putting dev names long time back.

api/src/main/resources/messages.properties
```diff
@@ -6,3 +6,10 @@ bahmni.appointment.teleConsultation.patientEmailNotificationTemplate=Hi {0}, You
 bahmni.appointment.adhocTeleConsultation.patientEmailNotificationSubject=Teleconsultation started
 bahmni.appointment.adhocTeleConsultation.patientEmailNotificationTemplate=Hi {0}, {1} has started a tele-consultation with you! Please use this link {3}. If you have any questions, please reach out to administration for assistance. See you soon!
 bahmni.appointment.adhocTeleConsultation.bccEmails=
+
+sms.appointmentReminderSMSTemplate=This is a reminder for upcoming Appointment for {0}:{1}\non Date:{2},with {3} for {4} at {5}.\nFor any queries call us on {6}
+sms.appointmentReminderSMSTemplateWithoutProvider=This is a reminder for upcoming Appointment for {0}:{1}\non Date:{2} for {3} at {4}.\nFor any queries call us on {5}
+
+bahmni.sms.timezone=IST
```
gsluthra, line 13: Why is configuration added in messages.properties?

api/src/main/resources/moduleApplicationContext.xml
```diff
@@ -83,7 +89,24 @@
...
     </bean>
+    <bean id="connectionDetails" class="org.openmrs.module.appointments.connection.ConnectionDetails" depends-on="openmrsProperties">
+        <property name="authUri" value="${openmrs.auth.uri}" />
+        <property name="openmrsUser" value="${openmrs.user}" />
+        <property name="openmrsPassword" value="${openmrs.password}"/>
+        <property name="openmrsWebclientConnectTimeout" value="${openmrs.connectionTimeoutInMilliseconds}" />
+        <property name="openmrsWebclientReadTimeout" value="${openmrs.replyTimeoutInMilliseconds}" />
+    </bean>
 
+    <bean id="appointmentVisitLoaction" class="org.openmrs.module.appointments.service.impl.AppointmentVisitLocation" depends-on="openmrsProperties">
```
gsluthra, line 100: typo in spelling of bean: `appointmentVisitLoaction`.

api/src/test/resources/TestingApplicationContext.xml
```diff
@@ -121,6 +121,14 @@
         <property name="appointmentDao" ref="appointmentDao"/>
     </bean>
 
+    <bean id="appointmentVisitLoaction" class="org.openmrs.module.appointments.service.impl.AppointmentVisitLocation" lazy-init="true" depends-on="openmrsProperties"/>
```
gsluthra, line 124: typo in bean id. Spelling of Location is incorrect.

api/src/test/resources/TestingApplicationContext.xml
```diff
@@ -121,6 +121,14 @@
         <property name="appointmentDao" ref="appointmentDao"/>
     </bean>
 
+    <bean id="appointmentVisitLoaction" class="org.openmrs.module.appointments.service.impl.AppointmentVisitLocation" lazy-init="true" depends-on="openmrsProperties"/>    
+    <bean id="openmrsLogin" class="org.openmrs.module.appointments.connection.OpenmrsLogin"/>
+
+    <bean id="smsService" class="org.openmrs.module.appointments.service.impl.SMSService" lazy-init="true">
+        <property name="openmrsLogin" ref="openmrsLogin"/>
+        <property name="appointmentVisitLocation" ref="appointmentVisitLoaction"/>
```
gsluthra, line 129: Typo. Spelling of `location` is incorrect.

omod/pom.xml
```diff
@@ -39,7 +39,7 @@
 		<dependency>
 			<groupId>org.bahmni.module</groupId>
 			<artifactId>teleconsultation-api</artifactId>
-			<version>1.0.0-SNAPSHOT</version>
+			<version>1.0.0</version>
```
gsluthra, line 42: Why are you removing SNAPSHOT? This is not a released module. Or is this a dependency, and thats why?

omod/src/main/resources/config.xml
```diff
@@ -90,6 +90,11 @@
 		<defaultValue>false</defaultValue>
 		<description>Scheduler marks scheduled appointments as missed when it is turned on</description>
 	</globalProperty>
+	<globalProperty>
+	<property>SchedulerReminderBeforeHours</property>
+	<defaultValue>24</defaultValue>
+	<description>Before how many hours the reminder is to be scheduled for appointment</description>
```
gsluthra, line 96: Rename this description to: `How many hours before the appointment should the reminder SMS be sent`

api/src/main/java/org/openmrs/module/appointments/service/impl/AppointmentsServiceImpl.java
atish160384: Sure will do that

api/src/main/java/org/openmrs/module/appointments/service/impl/SMSService.java
atish160384: Yah will do the change

api/src/main/resources/liquibase.xml
atish160384: Ohk was not aware of that will change to BAHMNI (just saw previous authors name :) so added

api/src/main/java/org/openmrs/module/appointments/service/impl/SMSService.java
atish160384: Gurpreet this condition was added as to check if the template in global property is empty or doesn't exist then it would evaluate the message using the template from message.properties file
Do we need to remove this condition and send empty message in that case
(which will remove's template from message.properties too)

omod/pom.xml
atish160384: This commit was added by @rahu1ramesh 
I guess he removed it because of that module being used as dependency can revert the same

api/src/main/java/org/openmrs/module/appointments/service/impl/SMSService.java
gsluthra: No. its okay. Let it be as-is then.

PR #121, opened 2023-08-23
Updated openmrs 3.x

mseaton:

PR #126, opened 2023-09-04
Bah 3201| Get appointments by date and status

ojwanganto:


api/src/main/java/org/openmrs/module/appointments/dao/impl/AppointmentDaoImpl.java
```diff
@@ -61,6 +61,25 @@ public List<Appointment> getAllAppointmentsReminder(String hours) {
...
+    public List<Appointment> getAllAppointments(Date forDate, String status) {
+        Criteria criteria = sessionFactory.getCurrentSession().createCriteria(Appointment.class);
+        criteria.add(Restrictions.eq("voided", false));
+        criteria.createAlias("patient", "patient");
+        criteria.add(Restrictions.eq("patient.voided", false));
+        criteria.add(Restrictions.eq("patient.personVoided", false));
+        if (forDate != null) {
+            Date maxDate = new Date(forDate.getTime() + TimeUnit.DAYS.toMillis(1));
+            criteria.add(Restrictions.ge("startDateTime", forDate));
+            criteria.add(Restrictions.lt("endDateTime", maxDate));
```
mseaton, line 74: There are a number of things here that seem wrong to me, but maybe I'm not fully understanding what this method aims to accomplish.

1. Your restrictions are not operating on the same column.  If the intention is to find all appointments that are optionally limited by date, I would think you'd want to only restrict on `startDatetime` (or at the very least you'd want to limit the `startDatetime` to both be within the passed date.
2. The way in which you are restricting the data also seems wrong.  Assuming what you want is all appointments on the given date, if someone passes in a `forDate` with a time component, the first criteria will not include appointments before the given time on that date.  And the second criteria will include appointments for the _next_ date, up until the time component of the original `forDate`.  You should first zero-out the time-component of `forDate` before using it.
3. The approach to get "one day later" also seems incorrect (by adding a day's worth of millis).  For example, what would happen on daylight savings dates where an hour is gained or lost?  You would not end up with the next day at the same time.  A better approach would be to use standard date/time functions (or Calendar) to add a day.

api/src/main/java/org/openmrs/module/appointments/dao/impl/AppointmentDaoImpl.java
```diff
@@ -61,6 +61,25 @@ public List<Appointment> getAllAppointmentsReminder(String hours) {
...
+        }
+
+        if (StringUtils.isNotEmpty(status)) {
+            criteria.add(Restrictions.eq("status", AppointmentStatus.valueOf(status))); //TODO: we may need to explore Optional construct to help validate against missing enum values
```
mseaton, line 78: See my earlier comment.  We should just have an enum passed into this method, not a String.  Validation can/should happen within the consumer (eg. the REST controller)

api/src/main/java/org/openmrs/module/appointments/dao/AppointmentDao.java
```diff
@@ -14,6 +14,7 @@ public interface AppointmentDao {
     List<Appointment> getAllAppointments(Date forDate);
     List<Appointment> getAllAppointmentsReminder(String afterTime);
 
+    List<Appointment> getAllAppointments(Date forDate, String status);
```
mseaton, line 17: As a service / DAO method, this should take in an actual AppointmentStatus enum, not a String representation of that enum. Consumers of this service should handle the conversion to/from enum and string and any validation needed.

api/src/main/java/org/openmrs/module/appointments/service/AppointmentsService.java
```diff
@@ -38,6 +38,9 @@ public interface AppointmentsService {
     @Authorized({VIEW_APPOINTMENTS, MANAGE_APPOINTMENTS})
     List<Appointment> getAllAppointmentsReminder(String hours);
 
+    @Transactional
+    @Authorized({VIEW_APPOINTMENTS, MANAGE_APPOINTMENTS})
+    List<Appointment> getAllAppointments(Date forDate, String status);
```
mseaton, line 43: See my earlier comment, this should take in an AppointmentStatus enum, not a String.

api/src/main/java/org/openmrs/module/appointments/service/AppointmentsService.java
```diff
@@ -38,6 +38,9 @@ public interface AppointmentsService {
     @Authorized({VIEW_APPOINTMENTS, MANAGE_APPOINTMENTS})
     List<Appointment> getAllAppointmentsReminder(String hours);
 
+    @Transactional
+    @Authorized({VIEW_APPOINTMENTS, MANAGE_APPOINTMENTS})
```
mseaton, line 42: Why does one need MANAGE_APPOINTMENTS privilege to get appointments?  This doesn't seem right, even if the other service methods already make this mistake.

api/src/main/java/org/openmrs/module/appointments/service/AppointmentsService.java
mseaton, line 43: There needs to be JavaDoc _somewhere_, even better in multiple places, that describes what this method does, especially given the implementation in the DAO.

omod/src/main/java/org/openmrs/module/appointments/web/controller/AppointmentController.java
```diff
@@ -191,4 +192,24 @@ public ResponseEntity<Object> rescheduleAppointment(@PathVariable("uuid") String
...
+     * @param status - the appointment status i.e. scheduled, cancelled, completed,
+     * @return list
+     */
+    @RequestMapping(method = RequestMethod.GET, value = "appointmentStatus")
+    @ResponseBody
+    public ResponseEntity<Object> getAppointmentByDateAndStatus(@RequestParam(value = "forDate") String forDate, @RequestParam(value = "status") String status)  {
+        try {
+            if(StringUtils.isEmpty(forDate) || StringUtils.isEmpty(status)) {
+                return new ResponseEntity<>("The request requires appointment date and status", HttpStatus.BAD_REQUEST);
+            }
```
mseaton, line 207: This is where you should validate both the date (parse it into a Date object) and the status (parse it into the enum).  If either fail, send an error response.  Then pass the date and the actual enum into the service.

api/src/main/java/org/openmrs/module/appointments/service/impl/AppointmentsServiceImpl.java
```diff
@@ -220,6 +228,13 @@ public List<Appointment> getAllAppointmentsReminder(String hours) {
 
     }
 
+    @Transactional
```
mseaton, line 231: This should be read-only Transactional(readOnly=true)

api/src/main/java/org/openmrs/module/appointments/service/impl/AppointmentsServiceImpl.java
```diff
@@ -220,6 +228,13 @@ public List<Appointment> getAllAppointmentsReminder(String hours) {
...
+    @Override
+    public List<Appointment> getAllAppointments(Date forDate, String status) {
```
mseaton, line 233: There should be unit tests for this that cover standard usage at the very least.

mseaton (changes requested):
@ojwanganto - please see my comments and changes requested.  @angshu who on the Bahmni team would typically review these and help shepherd them along?

angshu:
@ojwanganto also please sign the CLA

api/src/main/java/org/openmrs/module/appointments/service/AppointmentsService.java
angshu: MANAGE_APPOINTMENTS covers all - view as well as add/edit. At least that the impression I had.

mseaton: Oh, interesting.  By default, this is an _or_ operation.  I had just assumed this was an _and_.

ojwanganto:
> @ojwanganto also please sign the CLA

I may be doing something wrong that's making this not pick even after I sign. I have posted the same on slack

gsluthra:
The email associated with the commit needs to be the same as your Github email ID, which you used to sign the CLA.

api/src/main/java/org/openmrs/module/appointments/dao/impl/AppointmentDaoImpl.java
ojwanganto: @mseaton getAllAppointments is really not our addition. Please see https://github.com/Bahmni/openmrs-module-appointments/blob/master/api/src/main/java/org/openmrs/module/appointments/dao/impl/AppointmentDaoImpl.java#L34. The reviews are good and are worth addressing. They, however, should not block accepting the changes that were added by members outside of Bahmni team. I propose a comprehensive review of the module with a view of improving the legacy code if that is agreeable. @angshu what's your thoughts on this?

mseaton: @ojwanganto I see what you mean.  Still, it would probably be better if we could not duplicate this code across multiple DAO methods, as if it needs to be changed then it will need to be done consistently in multiple places.  Why not change the existing DAO to take in an optional status, and limit by this if it is non-null?  Then your PR makes clear that it is not introducing any code that has potentially questionable logic.

@angshu also interested in your opinion on my original comments on - what is existing functionality in the getAllAppointment method implementation.

omod/src/main/java/org/openmrs/module/appointments/web/controller/AppointmentController.java
mogoodrich: So one reason I think for passing in a string here instead of an enum is that, looking at the larger PR that was originally issued (https://github.com/Bahmni/openmrs-module-appointments/pull/121/files#diff-65f85cb43e49116cee3dce7e2e006a06cbaf9d83de10b6d4af933eab4ca75853R209) the intent is to allow searching on "statuses" that aren't part of the Status enum ("Pending" and "Honoured")... these states aren't defined by actual statuses stored on the appointment, but rather are defined by custom DAO methods, "getPendingAppointments", "getHonouredAppointments".  

I'm actually wondering if we actually need/want to add a getAppointmentByDateAndStatus endpoint and service method at all, or if  rather we should make custom endpoints for "pendingAppointments" and "honourAppoinments" like we have for "earlyAppointments" and "unScheduledAppointments"?  

Making "getAppointmentByDateAndStatus" search on both "real" enum statuses and derived "statuses" seems confused...

mogoodrich (commented):
Added a thought @CynthiaKamau @ojwanganto  @mseaton ... not to derail things, but if I'm understanding the current design, we may not need/want this new method at all?

ojwanganto:
> Added a thought @CynthiaKamau @ojwanganto @mseaton ... not to derail things, but if I'm understanding the current design, we may not need/want this new method at all?

This is very much in order. We are dropping honored appointments as it is really implementation specific and requires a generic hook for implementations to plug in their logic for this. I don't think it is relevant anymore.

mogoodrich:
> 
> This is very much in order. We are dropping honored appointments as it is really implementation specific and requires a generic hook for implementations to plug in their logic for this. I don't think it is relevant anymore.

That would be great if it works for you @ojwanganto ... especially because (you can correct me if I'm wrong) I think dateHonored is the only data model change you made, and I would think data model changes would be the most difficult to get approval for?

api/src/main/java/org/openmrs/module/appointments/dao/AppointmentDao.java
mogoodrich: Agree with @mseaton , this should take an actual AppointmentStatus enum instead of a string.

api/src/main/java/org/openmrs/module/appointments/dao/impl/AppointmentDaoImpl.java
mogoodrich: Agree with @mseaton about changing the existing service to take an optional status instead of creating a new method.  If we need to preserve the existing message signature, we can keep the existing signature but have it delegate to the one that takes status, ie:

```
public List<Appointment> getAllAppointments(Date forDate) {
    return getAllAppointments(forDate, null)
}
```

If we do this, we can merge, and we can address the concerns @mseaton had about the original method separately.
.

api/src/main/java/org/openmrs/module/appointments/dao/impl/AppointmentDaoImpl.java
mogoodrich: Agreed!

api/src/main/java/org/openmrs/module/appointments/service/AppointmentsService.java
mogoodrich: Agreed!

mogoodrich: Agree that we should add a javadoc here... it looks like there aren't any java docs on the other methods here, but let's get in the practice and add it for new methods we add or existing ones we modify.

api/src/main/java/org/openmrs/module/appointments/service/impl/AppointmentsServiceImpl.java
mogoodrich: Agreed, this is existing behavior, but we should change this to readOnly=true.  fyi @angshu

api/src/main/java/org/openmrs/module/appointments/service/impl/AppointmentsServiceImpl.java
mogoodrich: Agreed.  Looks like DAO tests can be added to the existing AppointmentDaoImplIT class, and there's some sample test data you can use, and then the service tests would go in AppointmentServiceTest class.

omod/src/main/java/org/openmrs/module/appointments/web/controller/AppointmentController.java
mogoodrich: So my above comment is a bit tangential to this... but, yes, if we want to merge this commit in as-is, as @mseaton said this is where we should parse into an enum,

mogoodrich (commented):
@ojwanganto  added my thoughts (tried to flag Cynthia what is her github handle?)... basically, I agree with all the points that @mseaton  raised.  And although there is a questoin for @angshu  about the existing implementatoin of the getAllAppointments, I don't think we need to get that resolved to get this merged in.

ojwanganto:
@mogoodrich @mseaton - there is a new PR issued by Cynthia. We had talked about closing this. I hope that is in order

mogoodrich:
Sounds good to me @ojwanganto !

PR #129, opened 2023-09-15
BAH-3223 - Appointment module logs errors at startup if atom feed mod…

mseaton:
…ule is not installed


atom-feed/src/main/java/org/openmrs/module/appointments/advice/AtomFeedAdviceProvider.java
```diff
@@ -0,0 +1,33 @@
...
+import org.openmrs.module.appointments.service.AppointmentsService;
+import org.springframework.beans.BeansException;
+import org.springframework.context.ApplicationContext;
+import org.springframework.context.ApplicationContextAware;
+
+/**
+ * Registers AOP advice when the context starts up.
+ * This is done instead of declaring advice in config.xml in order to allow it to be conditionally loaded
+ */
+public class AtomFeedAdviceProvider implements ApplicationContextAware {
```
angshu, line 18: How is this wired up? through component scan? 
Also should the advices be wired up conditionally, since the intention is to work without the atomfeed module? should @Conditional annotation be used?

ojwanganto: Is there a preferred approach to implementing such conditional loading? I was hoping this had been resolved so that i learn from it

omod/src/test/java/org/openmrs/module/appointments/web/controller/AppointmentControllerTest.java
```diff
@@ -1,5 +1,6 @@
 package org.openmrs.module.appointments.web.controller;
```
mseaton, line 1: The changes in this file are not for this PR, but pending in another PR (see BAH-3222).  I need these in place to successfully build and test.

api/src/test/java/org/openmrs/module/appointments/service/impl/DefaultTeleconsultationAppointmentPatientEmailNotifierTest.java
```diff
@@ -31,6 +32,7 @@
 
 @RunWith(PowerMockRunner.class)
 @PrepareForTest(Context.class)
+@PowerMockIgnore( {"javax.*", "org.apache.*", "org.slf4j.*"} )
```
mseaton, line 35: This is just an incidental fix- this was throwing stack traces in the logs during the build, and adding this ignore line fixes that.

api/src/main/java/org/openmrs/module/appointments/AppointmentsActivator.java
```diff
@@ -27,18 +30,34 @@ public class AppointmentsActivator extends BaseModuleActivator {
 	private Log log = LogFactory.getLog(this.getClass());
```
mseaton, line 31: I moved away from the original approach of using ApplicationContextAware, in favor of a more generally understood approach of using the module activator and iterating over registered components.  I made this generalized as it is not specific to advice, but could be utilized for other reasons if needed.

One of the main reasons I moved to this approach is to make an attempt to also clean up / remove advice if the module is stopped like is done with standard advice in the ModuleFactory.  This is just an additional precaution, though OpenMRS doesn't really function properly anyway (or I wouldn't trust it to) if modules are started/stopped via the legacy admin interface, and this isn't really a huge concern.

atom-feed/src/main/java/org/openmrs/module/appointments/advice/AtomFeedAdviceActivatorComponent.java
```diff
@@ -0,0 +1,63 @@
...
+/**
+ * Registers AOP advice when the context starts up.
+ * This is done instead of declaring advice in config.xml in order to allow it to be conditionally loaded
+ */
+@OpenmrsProfile(modules = { "openmrs-atomfeed:*" })
+public class AtomFeedAdviceActivatorComponent implements AppointmentsActivatorComponent {
+
+    private static final Log log = LogFactory.getLog(AtomFeedAdviceActivatorComponent.class);
+
+    private final AppointmentServiceDefinitionAdvice appointmentServiceDefinitionAdvice;
```
mseaton, line 22: I needed to use these properties and two different constructors just so that I could properly mock in unit tests.

mseaton (commented):
@angshu / @gsluthra - see my comments here and on the ticket.  Would be good if you could have a look over this.

rahu1ramesh (approved):

angshu (approved):

PR #133, opened 2023-09-27
BAH-3201 Add a method to get appointments by date and status

CynthiaKamau:
This is a follow up PR for https://github.com/Bahmni/openmrs-module-appointments/pull/126
Get appointments for a specific date and status is an endpoint used in the O3 ref app appointments homepage

### DIFF
--- a/api/src/main/java/org/openmrs/module/appointments/dao/AppointmentDao.java
+++ b/api/src/main/java/org/openmrs/module/appointments/dao/AppointmentDao.java
@@ -12,6 +12,9 @@ import java.util.List;
 
 public interface AppointmentDao {
     List<Appointment> getAllAppointments(Date forDate);
+
+    List<Appointment> getAllAppointments(Date forDate, AppointmentStatus status);
+
     List<Appointment> getAllAppointmentsReminder(String afterTime);
 
     @Transactional
--- a/api/src/main/java/org/openmrs/module/appointments/dao/AppointmentDao.java
+++ b/api/src/main/java/org/openmrs/module/appointments/dao/AppointmentDao.java
@@ -12,9 +12,6 @@ import java.util.List;
 
 public interface AppointmentDao {
     List<Appointment> getAllAppointments(Date forDate);
-
-    List<Appointment> getAllAppointments(Date forDate, AppointmentStatus status);
-
     List<Appointment> getAllAppointmentsReminder(String afterTime);
 
     @Transactional
--- a/api/src/main/java/org/openmrs/module/appointments/dao/impl/AppointmentDaoImpl.java
+++ b/api/src/main/java/org/openmrs/module/appointments/dao/impl/AppointmentDaoImpl.java
@@ -19,6 +19,7 @@ import org.springframework.transaction.annotation.Transactional;
 
 import java.util.Date;
 import java.util.List;
+import java.util.Arrays;
 import java.util.concurrent.TimeUnit;
 
 public class AppointmentDaoImpl implements AppointmentDao {
@@ -32,17 +33,7 @@ public class AppointmentDaoImpl implements AppointmentDao {
 
     @Override
     public List<Appointment> getAllAppointments(Date forDate) {
-        Criteria criteria = sessionFactory.getCurrentSession().createCriteria(Appointment.class);
-        criteria.add(Restrictions.eq("voided", false));
-        criteria.createAlias("patient", "patient");
-        criteria.add(Restrictions.eq("patient.voided", false));
-        criteria.add(Restrictions.eq("patient.personVoided", false));
-        if (forDate != null) {
-            Date maxDate = new Date(forDate.getTime() + TimeUnit.DAYS.toMillis(1));
-            criteria.add(Restrictions.ge("startDateTime", forDate));
-            criteria.add(Restrictions.lt("endDateTime", maxDate));
-        }
-        return criteria.list();
+        return this.getAllAppointments(forDate, null);
     }
 
     @Override
@@ -58,6 +49,32 @@ public class AppointmentDaoImpl implements AppointmentDao {
             criteria.add(Restrictions.lt("startDateTime", maxDate));
         }
         criteria.add(Restrictions.ne("status", AppointmentStatus.Cancelled));
+        return criteria.list();
+    }
+
+    /**
+     * Retrieves a list of appointments based on the specified date and status criteria.
+     *
+     * @param forDate The date for filtering appointments. Null to ignore date filtering.
+     * @param status The status for filtering appointments. Null to ignore status filtering.
+     * @return A list of matching appointments, possibly empty.
+     */
+    @Override
+    public List<Appointment> getAllAppointments(Date forDate, AppointmentStatus status) {
+        Criteria criteria = sessionFactory.getCurrentSession().createCriteria(Appointment.class);
+        criteria.add(Restrictions.eq("voided", false));
+        criteria.createAlias("patient", "patient");
+        criteria.add(Restrictions.eq("patient.voided", false));
+        criteria.add(Restrictions.eq("patient.personVoided", false));
+        if (forDate != null) {
+            Date maxDate = new Date(forDate.getTime() + TimeUnit.DAYS.toMillis(1));
+            criteria.add(Restrictions.ge("startDateTime", forDate));
+            criteria.add(Restrictions.lt("endDateTime", maxDate));
+        }
+
+        if (status != null) {
+            criteria.add(Restrictions.eq("status", status));
+        }
         return criteria.list();
     }
 
--- a/api/src/main/java/org/openmrs/module/appointments/dao/impl/AppointmentDaoImpl.java
+++ b/api/src/main/java/org/openmrs/module/appointments/dao/impl/AppointmentDaoImpl.java
@@ -19,7 +19,6 @@ import org.springframework.transaction.annotation.Transactional;
 
 import java.util.Date;
 import java.util.List;
-import java.util.Arrays;
 import java.util.concurrent.TimeUnit;
 
 public class AppointmentDaoImpl implements AppointmentDao {
@@ -33,7 +32,17 @@ public class AppointmentDaoImpl implements AppointmentDao {
 
     @Override
     public List<Appointment> getAllAppointments(Date forDate) {
-        return this.getAllAppointments(forDate, null);
+        Criteria criteria = sessionFactory.getCurrentSession().createCriteria(Appointment.class);
+        criteria.add(Restrictions.eq("voided", false));
+        criteria.createAlias("patient", "patient");
+        criteria.add(Restrictions.eq("patient.voided", false));
+        criteria.add(Restrictions.eq("patient.personVoided", false));
+        if (forDate != null) {
+            Date maxDate = new Date(forDate.getTime() + TimeUnit.DAYS.toMillis(1));
+            criteria.add(Restrictions.ge("startDateTime", forDate));
+            criteria.add(Restrictions.lt("endDateTime", maxDate));
+        }
+        return criteria.list();
     }
 
     @Override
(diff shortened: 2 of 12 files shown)


api/src/main/java/org/openmrs/module/appointments/model/AppointmentStatus.java
```diff
@@ -1,7 +1,7 @@
 package org.openmrs.module.appointments.model;
 
 public enum AppointmentStatus {
-    Requested("Requested", 0), Scheduled("Scheduled", 1), CheckedIn("CheckedIn", 2), Completed("Completed", 3), Cancelled("Cancelled", 3), Missed("Missed", 3);
+    Requested("Requested", 0), Scheduled("Scheduled", 1), CheckedIn("CheckedIn", 2), Completed("Completed", 3), Cancelled("Cancelled", 4), Missed("Missed", 5), CameEarly("CameEarly", 6), Pending("Pending", 7) ;
```
ojwanganto, line 4: Could you expound on the **_pending_** appointment status?

api/src/main/java/org/openmrs/module/appointments/dao/impl/AppointmentDaoImpl.java
```diff
@@ -61,6 +52,30 @@ public List<Appointment> getAllAppointmentsReminder(String hours) {
         return criteria.list();
     }
 
+    @Override
+    public List<Appointment> getAllAppointments(Date forDate, AppointmentStatus status) {
```
ojwanganto, line 56: Let's add javadoc to describe what the method does.

ojwanganto (changes requested):

ojwanganto:
@mseaton kindly help with reviews on the changes

ojwanganto:
> This is a follow up PR for #126 Get appointments for a specific date and status is an endpoint used in the O3 ref app appointments homepage

Please remember to sign the CLA

omod/src/main/java/org/openmrs/module/appointments/web/controller/AppointmentController.java
```diff
@@ -191,4 +192,33 @@ public ResponseEntity<Object> rescheduleAppointment(@PathVariable("uuid") String
...
+            } catch (IllegalArgumentException e) {
+                return new ResponseEntity<>("Invalid status value", HttpStatus.BAD_REQUEST);
+            }
+            List<Appointment> appointments = appointmentsService.getAllAppointments(date, appointmentStatus);
+            return new ResponseEntity<>(appointmentMapper.constructResponse(appointments), HttpStatus.OK);
+        } catch (Exception e) {
+            log.error("Runtime error while trying to fetch appointments by status and date", e);
+            return new ResponseEntity<>(RestUtil.wrapErrorResponse(e, e.getMessage()), HttpStatus.INTERNAL_SERVER_ERROR);
+        }
+    }
```
mseaton, line 223: I don't think this REST endpoint is what we want.  Having a new endpoint called "appointmentStatus" does not feel like the right verbage.  What you really want to do (I think) is just to use the existing "all" endpoint, but modify it to filter on status as well as on date (much like you did in the service and dao layers).  So, can't we - instead of creating a new endpoint here - simply modify the `getAllAppointments` method that already exists to take in an optional `status` request parameter, parses this status parameter into an enum (as you are doing above) and then call the new service method you just created that takes in both the date and the status?

mseaton (changes requested):
See comments throughout the PR.  I think this approach to adding a new service method is generally fine.  I will mention 2 things though:

1. It isn't 100% clear to me how much value this new service/dao/rest endpoint are actually going to add vs. just getting all appointments and filtering on the client side.

2. There is an existing "search" REST endpoint that already allows one to get appointments with a given status.  The problem with it is that it doesn't allow filtering by date - due largely I think to the fact that the underlying service search method uses an "Example" appointment to express the search criteria, and the "date" upon which one wants to search is typically more nuanced than that - eg. you don't want to get all appointments on a specific datetime, you want to get all appointments in a datetime range (like all on a particular date, regardless of time).  I could see there being reasonable value in refactoring and fixing (potentially deprecating the existing search method in favor of a new search method that takes in an `AppointmentSearchCriteria` rather than an example Appointment.  This would then allow for  broad search capabilities of appointments, using a single, standard `search` REST endpoint that returned a standard set of response objects.

Things to consider...

ojwanganto:
> See comments throughout the PR. I think this approach to adding a new service method is generally fine. I will mention 2 things though:
_**It isn't 100% clear to me how much value this new service/dao/rest endpoint are actually going to add vs. just getting all appointments and filtering on the client side.**_ - Not all requests for appointments on a date should be subjected to massive filtering on the frontend. Even though we may have alternative approaches for the current UI needs, we definitely need to keep this method as it provides for granular searching of appointments. What risks do you see with this?

api/src/main/java/org/openmrs/module/appointments/dao/impl/AppointmentDaoImpl.java
```diff
@@ -61,6 +52,30 @@ public List<Appointment> getAllAppointmentsReminder(String hours) {
...
+        criteria.add(Restrictions.eq("patient.voided", false));
+        criteria.add(Restrictions.eq("patient.personVoided", false));
+        if (forDate != null) {
+            Date maxDate = new Date(forDate.getTime() + TimeUnit.DAYS.toMillis(1));
+            criteria.add(Restrictions.ge("startDateTime", forDate));
+            criteria.add(Restrictions.lt("endDateTime", maxDate));
+        }
+
+        if (status != null) {
+            boolean isValidStatus = Arrays.asList(AppointmentStatus.values()).contains(status);
```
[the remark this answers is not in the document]
CynthiaKamau: updated this

mseaton:
> What risks do you see with this?

None @ojwanganto , you're right.

api/src/main/java/org/openmrs/module/appointments/dao/AppointmentDao.java
```diff
@@ -12,6 +12,8 @@
 
 public interface AppointmentDao {
     List<Appointment> getAllAppointments(Date forDate);
+    List<Appointment> getAllAppointments(Date forDate, AppointmentStatus status);
```
mseaton, line 15: Add a new line between these, just so that this is consistent with the rest of the file

api/src/main/java/org/openmrs/module/appointments/dao/impl/AppointmentDaoImpl.java
mseaton: We can mark this as resolved.

api/src/main/java/org/openmrs/module/appointments/service/impl/AppointmentsServiceImpl.java
```diff
@@ -220,6 +228,13 @@ public List<Appointment> getAllAppointmentsReminder(String hours) {
 
     }
 
+    @Transactional
```
mseaton, line 231: I personally think this should be `@Transactional(readOnly = true)`, but I'm fine just doing what you are doing to be consistent with the rest of the file I guess, and then we can create a new ticket to clean all of this up.

I do think that, since you have this new service impl method, that you should change the existing `getAllAppointments(Date)` method to delegate to this one, just like you did in the DAO.

omod/src/main/java/org/openmrs/module/appointments/web/controller/AppointmentController.java
```diff
@@ -190,5 +191,4 @@ public ResponseEntity<Object> rescheduleAppointment(@PathVariable("uuid") String
             return new ResponseEntity<>(RestUtil.wrapErrorResponse(e, e.getMessage()), HttpStatus.BAD_REQUEST);
         }
     }
-
 }
```
mseaton, line 194: Seems like a few changes coming through in this file that were unintentional?

mseaton (changes requested):
A few very minor comments that should be quick to address @CynthiaKamau .  Thanks for working through this.  I assume there will be another PR after this that supports the REST endpoint?

mseaton (changes requested):
Looking close @CynthiaKamau .  But we do need unit tests for this.  You should be able to just look at what existing unit tests exist for the `getAllAppointments(Date)` endpoint, and add similar tests for `getAllAppointments(Date, AppointmentStatus)`.

api/src/main/java/org/openmrs/module/appointments/model/AppointmentStatus.java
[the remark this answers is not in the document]
mogoodrich: Looks like this has already been removed, but I 100% agree with pulling this out for now.

When we do get back to this, we should make sure anything we add matches (or can be mapped to) one of standard FHIR statuses unless we have a strong reason not to:

https://build.fhir.org/valueset-appointmentstatus.html

(And it seems like this should be discussed in a ticket rather in a PR)

CynthiaKamau:
> Looking close @CynthiaKamau . But we do need unit tests for this. You should be able to just look at what existing unit tests exist for the `getAllAppointments(Date)` endpoint, and add similar tests for `getAllAppointments(Date, AppointmentStatus)`.

I have added some tests, kindly review them.

mseaton (approved):
This looks good to me @CynthiaKamau , Nice work and thanks for working through this.  It looks like you still have not signed the CLA, so this won't be able to be merged until you do.  Can you please do that?

ojwanganto:
> This looks good to me @CynthiaKamau , Nice work and thanks for working through this. It looks like you still have not signed the CLA, so this won't be able to be merged until you do. Can you please do that?

There could be something else that is required for CLA signing. I witnessed her sign but still it couldn't reflect.

mogoodrich:
Thoughts @angshu ?  For what it's worth, I was just able to sign the CLA (but I guess without entering a PR I can't confirm that it works for me)

mseaton:
> There could be something else that is required for CLA signing. I witnessed her sign but still it couldn't reflect

As @angshu indicated the other week, one needs to sign the CLA using the same email as what is associated with one's github account.  Please confirm that this is what is happening.

CynthiaKamau:
> > There could be something else that is required for CLA signing. I witnessed her sign but still it couldn't reflect
> 
> As @gsluthra indicated last week [in this PR](https://github.com/Bahmni/openmrs-module-appointments/pull/126#issuecomment-1727701931), one needs to sign the CLA using the same email as what is associated with one's github account. Please confirm that this is what is happening.

I have managed to sign it

mseaton:
Not sure who the list of maintainers is for this repo, but @gsluthra / @angshu - I think this is ready for your review and ideally to merge in.

gsluthra:
I have requested @angshu to review/approve.

angshu:
While this PR is ok, I wonder why introduce this here? there is already an API 
- /appointment/search  - that takes a date and status
- /appointments/search - does not take status
Note, the /appointment/search resides in AppointmentController, which was to be reimplemented. So will also encourage moving to the new AppointmentsController.

mogoodrich:
Good point @angshu ...  @CynthiaKamau @ojwanganto  can you confirm that this search can be performed with the POST "appointment/search" endpoint, or is there something lacking there?

(@angshu to you point, we'll have to us the old appointment/search functionality for now until we implement status search in the new(er) appointments/search endpoint)

CynthiaKamau:
> Good point @angshu ... @CynthiaKamau @ojwanganto can you confirm that this search can be performed with the POST "appointment/search" endpoint, or is there something lacking there?
> 
> (@angshu to you point, we'll have to us the old appointment/search functionality for now until we implement status search in the new(er) appointments/search endpoint)



> Good point @angshu ... @CynthiaKamau @ojwanganto can you confirm that this search can be performed with the POST "appointment/search" endpoint, or is there something lacking there?
> 
> (@angshu to you point, we'll have to us the old appointment/search functionality for now until we implement status search in the new(er) appointments/search endpoint)

The search can be performed by `/appointments/search` but i will have to add status as one of the params

mogoodrich (commented):
Thanks @CynthiaKamau !  Great to simplify this and use existing functionality and extend it to do what we need.  Can you just add one or two tests to the AppointmentsControllerIT just to confirm this works properly and then from my end this would be ready to merge: https://github.com/Bahmni/openmrs-module-appointments/blob/master/omod/src/test/java/org/openmrs/module/appointments/web/controller/AppointmentsControllerIT.java#L116

CynthiaKamau:
> Thanks @CynthiaKamau ! Great to simplify this and use existing functionality and extend it to do what we need. Can you just add one or two tests to the AppointmentsControllerIT just to confirm this works properly and then from my end this would be ready to merge: https://github.com/Bahmni/openmrs-module-appointments/blob/master/omod/src/test/java/org/openmrs/module/appointments/web/controller/AppointmentsControllerIT.java#L116

Done

mogoodrich (approved):
LGTM, thanks @CynthiaKamau !

@angshu @gsluthra  can we get this approved and merged?  Thanks!

angshu (approved):

PR #136, opened 2023-10-05
BAH-3249 - Change javax.mail from compile to test dependency

mseaton:


mogoodrich:
LGTM, with my limited knowledge!

mogoodrich:
Thanks @gsluthra !

PR #117, opened 2023-07-07
BAH 2962 | SMS functionality of appointment booking and reminder moved to backend

atish160384:
SMS functionality of appointment booking and reminder moved to backend

PR #141, opened 2023-11-21
AppointmentEventsAdvice should not throw when older validateAndSave is called

ibacher:
Apologies for not creating a ticket; my browser and Atlassian cloud are not working at the moment.

In #117, `AppointmentEventsAdvice` was added to enable publishing events when appointments are created. This advice is bound to the `validateAndSave()` method on `AppointmentService`. Unfortunately, there are _two_ `validateAndSave()` methods on `AppoinmentService`, one which takes a `Supplier<Appointment>` and one which just takes an appointment. The advice assumes it is called by the version that takes a `Supplier<Appointment>` so that if the other one is called, an exception is thrown, aborting the attempt to save.

This PR fixes that by ensuring the advice also works if the first argument is not an appointment. I also added some extra defensive checks to ensure that the `Supplier` returns an `Appointment` as expected.

### DIFF
--- a/api/src/main/java/org/openmrs/module/appointments/events/advice/AppointmentEventsAdvice.java
+++ b/api/src/main/java/org/openmrs/module/appointments/events/advice/AppointmentEventsAdvice.java
@@ -4,13 +4,13 @@ import com.google.common.collect.Sets;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Logger;
 import org.openmrs.api.context.Context;
-import org.openmrs.module.appointments.events.AppointmentEvent;
 import org.openmrs.module.appointments.events.publisher.AppointmentEventPublisher;
 import org.openmrs.module.appointments.model.Appointment;
 import org.openmrs.module.appointments.events.AppointmentBookingEvent;
 import org.openmrs.module.appointments.events.AppointmentEventType;
 import org.springframework.aop.AfterReturningAdvice;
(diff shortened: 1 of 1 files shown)


ojwanganto:
@angshu this PR is very critical. We encountered the same error and had to customize the code to support our implementations. I will appreciate if we get someone look into this

gokultw:
> In #117, `AppointmentEventsAdvice` was added to enable publishing events when appointments are created. This advice is bound to the `validateAndSave()` method on `AppointmentService`. Unfortunately, there are _two_ `validateAndSave()` methods on `AppoinmentService`, one which takes a `Supplier<Appointment>` and one which just takes an appointment. The advice assumes it is called by the version that takes a `Supplier<Appointment>` so that if the other one is called, an exception is thrown, aborting the attempt to save.
> 
> This PR fixes that by ensuring the advice also works if the first argument is not an appointment. I also added some extra defensive checks to ensure that the `Supplier` returns an `Appointment` as expected.

@ojwanganto, I have created a card https://bahmni.atlassian.net/browse/BAH-3854, someone who has bandwidth will pick this up soon.

rahu1ramesh (approved):

rahu1ramesh:
@ibacher @ojwanganto I have approved and merged the changes from the PR. Could you please validate if the issue is resolved? Thank you!

ojwanganto:
Thanks @rahu1ramesh

PR #142, opened 2023-12-01
BAH-3350: Appointments: search endpoint stack traces when a patient h…

mogoodrich:
…as more than one identifier of the same time


binduak (approved):

kavitha-sundararajan (approved):

angshu:
@mogoodrich I am not sure if we should send out comma separated string that way. Ideally we should send out an array  however doing that may break existing UI. Sending a comma separated identifiers, may also break if client relies on identifier (and not UUID) to subsequently make API calls. but Thats prob not common. I will create a ticket for this.

PR #145, opened 2024-01-19
BAH-3479 | Backend changes for dateless appointments

kavitha-sundararajan:
JIRA - [BAH-3479](https://bahmni.atlassian.net/browse/BAH-3479)

### DIFF
--- a/api/src/main/java/org/openmrs/module/appointments/dao/impl/AppointmentDaoImpl.java
+++ b/api/src/main/java/org/openmrs/module/appointments/dao/impl/AppointmentDaoImpl.java
@@ -8,7 +8,6 @@ import org.hibernate.criterion.Example;
 import org.hibernate.criterion.Order;
 import org.hibernate.criterion.Restrictions;
 import org.hibernate.sql.JoinType;
-import org.openmrs.api.context.Context;
 import org.openmrs.module.appointments.dao.AppointmentDao;
 import org.openmrs.module.appointments.model.Appointment;
 import org.openmrs.module.appointments.model.AppointmentSearchRequest;
@@ -163,7 +162,8 @@ public class AppointmentDaoImpl implements AppointmentDao {
         criteria.add(Restrictions.eq("voided", false));
         if (appointmentSearchRequest.getStartDate() != null) {
             criteria.addOrder(Order.asc("startDateTime"));
-        } else {
+        }
+        if (appointmentSearchRequest.getStatus() != null && appointmentSearchRequest.getStatus().name().equals(AppointmentStatus.WaitList.name())) {
             criteria.addOrder(Order.asc("dateCreated"));
         }
         setDateCriteria(appointmentSearchRequest, criteria);
@@ -172,9 +172,6 @@ public class AppointmentDaoImpl implements AppointmentDao {
         setProviderCriteria(appointmentSearchRequest, criteria);
         setStatusCriteria(appointmentSearchRequest, criteria);
 
-        String limit = Context.getAdministrationService().getGlobalProperty("webservices.rest.maxResultsDefault");
-        if(StringUtils.isNotEmpty(limit))
-            criteria.setMaxResults(Integer.parseInt(limit));
         return criteria.list();
     }
 
@@ -233,10 +230,17 @@ public class AppointmentDaoImpl implements AppointmentDao {
     }
 
     @Override
-    public List<Appointment> getDatelessAppointments() {
+    public List<Appointment> getAppointmentsWithoutDates(Integer limit) {
         Criteria criteria = sessionFactory.getCurrentSession().createCriteria(Appointment.class);
+        criteria.createAlias("patient", "patient");
+        criteria.add(Restrictions.eq("patient.voided", false));
+        criteria.add(Restrictions.eq("patient.personVoided", false));
         criteria.add(Restrictions.isNull("startDateTime"));
         criteria.add(Restrictions.isNull("endDateTime"));
+        criteria.addOrder(Order.asc("dateCreated"));
+        if (limit != null) {
+            criteria.setMaxResults(limit);
+        }
         return criteria.list();
     }
 }
--- a/api/src/main/java/org/openmrs/module/appointments/dao/impl/AppointmentDaoImpl.java
+++ b/api/src/main/java/org/openmrs/module/appointments/dao/impl/AppointmentDaoImpl.java
@@ -8,6 +8,7 @@ import org.hibernate.criterion.Example;
 import org.hibernate.criterion.Order;
 import org.hibernate.criterion.Restrictions;
 import org.hibernate.sql.JoinType;
+import org.openmrs.api.context.Context;
 import org.openmrs.module.appointments.dao.AppointmentDao;
 import org.openmrs.module.appointments.model.Appointment;
 import org.openmrs.module.appointments.model.AppointmentSearchRequest;
@@ -160,13 +161,20 @@ public class AppointmentDaoImpl implements AppointmentDao {
         Criteria criteria = sessionFactory.getCurrentSession().createCriteria(Appointment.class);
 
         criteria.add(Restrictions.eq("voided", false));
-        criteria.addOrder(Order.asc("startDateTime"));
+        if (appointmentSearchRequest.getStartDate() != null) {
+            criteria.addOrder(Order.asc("startDateTime"));
+        } else {
+            criteria.addOrder(Order.asc("dateCreated"));
+        }
         setDateCriteria(appointmentSearchRequest, criteria);
         setPatientCriteria(appointmentSearchRequest, criteria);
         setLimitCriteria(appointmentSearchRequest, criteria);
         setProviderCriteria(appointmentSearchRequest, criteria);
         setStatusCriteria(appointmentSearchRequest, criteria);
 
+        String limit = Context.getAdministrationService().getGlobalProperty("webservices.rest.maxResultsDefault");
+        if(StringUtils.isNotEmpty(limit))
+            criteria.setMaxResults(Integer.parseInt(limit));
         return criteria.list();
     }
 
--- a/api/src/main/java/org/openmrs/module/appointments/dao/impl/AppointmentDaoImpl.java
+++ b/api/src/main/java/org/openmrs/module/appointments/dao/impl/AppointmentDaoImpl.java
@@ -223,4 +223,12 @@ public class AppointmentDaoImpl implements AppointmentDao {
 
         return criteria.list();
     }
+
+    @Override
+    public List<Appointment> getDatelessAppointments() {
+        Criteria criteria = sessionFactory.getCurrentSession().createCriteria(Appointment.class);
+        criteria.add(Restrictions.isNull("startDateTime"));
+        criteria.add(Restrictions.isNull("endDateTime"));
+        return criteria.list();
+    }
 }
--- a/api/src/main/java/org/openmrs/module/appointments/dao/impl/AppointmentDaoImpl.java
+++ b/api/src/main/java/org/openmrs/module/appointments/dao/impl/AppointmentDaoImpl.java
@@ -160,12 +160,7 @@ public class AppointmentDaoImpl implements AppointmentDao {
         Criteria criteria = sessionFactory.getCurrentSession().createCriteria(Appointment.class);
 
         criteria.add(Restrictions.eq("voided", false));
-        if (appointmentSearchRequest.getStartDate() != null) {
-            criteria.addOrder(Order.asc("startDateTime"));
-        }
-        if (appointmentSearchRequest.getStatus() != null && appointmentSearchRequest.getStatus().name().equals(AppointmentStatus.WaitList.name())) {
-            criteria.addOrder(Order.asc("dateCreated"));
-        }
+        criteria.addOrder(Order.asc("startDateTime"));
         setDateCriteria(appointmentSearchRequest, criteria);
         setPatientCriteria(appointmentSearchRequest, criteria);
         setLimitCriteria(appointmentSearchRequest, criteria);
--- a/api/src/main/java/org/openmrs/module/appointments/helper/AppointmentServiceHelper.java
+++ b/api/src/main/java/org/openmrs/module/appointments/helper/AppointmentServiceHelper.java
@@ -51,9 +51,9 @@ public class AppointmentServiceHelper {
         appointmentJson.put("providerUuid", providerUuid);
         String locationUuid = appointment.getLocation() != null ? appointment.getLocation().getUuid() : null;
         appointmentJson.put("locationUuid", locationUuid);
-        String startDate = appointment.getStartDateTime() != null ? appointment.getStartDateTime().toInstant().toString() : null;
+        String startDate = (appointment.getStatus() == AppointmentStatus.WaitList && appointment.getStartDateTime() == null) ? null : appointment.getStartDateTime().toInstant().toString();
         appointmentJson.put("startDateTime", startDate);
-        String endDate = appointment.getEndDateTime() != null ? appointment.getEndDateTime().toInstant().toString() : null;
+        String endDate = (appointment.getStatus() == AppointmentStatus.WaitList && appointment.getStartDateTime() == null) ? null : appointment.getEndDateTime().toInstant().toString();
         appointmentJson.put("endDateTime", endDate);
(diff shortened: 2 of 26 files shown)


gsluthra:
As @angshu had mentioned yesterday, please ensure that an appointment should show in "Waitlist" based on status and not based on whether a date is set/no step. Can you verify code for this, and maybe add a test to ensure that even if a data is set on an appointment, it shows in waitlist if its status is waitlist.

api/src/test/java/org/openmrs/module/appointments/dao/impl/AppointmentDaoImplIT.java
```diff
@@ -278,29 +260,11 @@ public void shouldReturnEmptyListWhenPatientIsNull() {
         assertNotNull(appointments);
         assertEquals(0, appointments.size());
     }
-    @Test
```
gsluthra, line 281: Why are these tests removed?

omod/src/main/java/org/openmrs/module/appointments/web/contract/AppointmentRequest.java
```diff
@@ -126,4 +127,12 @@ public List<AppointmentProviderDetail> getProviders() {
     public void setProviders(List<AppointmentProviderDetail> providers) {
         this.providers = providers;
     }
+
+    public String getPriority() {
+        return priority;
+    }
+
+    public void setPriority(String priority) {
```
gsluthra, line 135: Why are getters/setters for Appointment Priority using String, instead of Enum?

omod/src/main/java/org/openmrs/module/appointments/web/contract/AppointmentDefaultResponse.java
```diff
@@ -176,4 +178,20 @@ public void setTeleconsultationLink(String teleconsultationLink) {
 	public String getTeleconsultationLink() {
 		return teleconsultationLink;
 	}
+
+	public void setPriority(String priority) {
```
gsluthra, line 182: Why are getters/setters for Appointment Priority using String, instead of Enum?

omod/src/test/java/org/openmrs/module/appointments/web/mapper/AppointmentMapperTest.java
```diff
@@ -229,6 +231,22 @@ public void shouldMapExistingAppointmentWhenPayloadHasAppointmentStatus() throws
 
     }
 
+    @Test
+    public void shouldMapExistingAppointmentWhenPayloadHasAppointmentPriority() throws ParseException {
+        AppointmentRequest appointmentRequest = createAppointmentRequest();
+        appointmentRequest.setPriority("Routine");
```
gsluthra, line 237: Why are getters/setters for Appointment Priority using String, instead of Enum?

omod/src/main/java/org/openmrs/module/appointments/web/contract/AppointmentRequest.java
kavitha-sundararajan: This is request contract class and we followed the same as that of appointment status

omod/src/main/java/org/openmrs/module/appointments/web/contract/AppointmentDefaultResponse.java
kavitha-sundararajan: This is response contract class and we followed the same as that of appointment status

omod/src/test/java/org/openmrs/module/appointments/web/mapper/AppointmentMapperTest.java
kavitha-sundararajan: Since we referred appointment status, we followed the same for priority as well

api/src/main/resources/liquibase.xml
```diff
@@ -671,4 +671,29 @@
...
+    </changeSet>
+    <changeSet id="remove-not-null-constraint-202304211826" author="Kavitha, Umair">
+        <preConditions onFail="MARK_RAN">
+            <tableExists tableName="patient_appointment" />
+            <columnExists tableName="patient_appointment" columnName="start_date_time" />
+            <columnExists tableName="patient_appointment" columnName="end_date_time" />
+        </preConditions>
+        <comment>Update start and end date time of patient appointment table to be nullable</comment>
+        <sql>
+            ALTER TABLE patient_appointment MODIFY start_date_time datetime;
```
gsluthra, line 695: I think its better to be explicit about default as null using `DEFAULT NULL`.

https://fedingo.com/how-to-modify-mysql-column-to-allow-null/

gsluthra:
What is the default priority assumed, if no priority is set (for backward compatibility)? Where is that code? I am guessing default is "routine" priority. thanks!

omod/src/main/java/org/openmrs/module/appointments/web/controller/AppointmentController.java
```diff
@@ -60,6 +60,10 @@ public List<AppointmentDefaultResponse> getAllAppointments(@RequestParam(value =
     @RequestMapping( method = RequestMethod.POST, value = "search")
     @ResponseBody
     public List<AppointmentDefaultResponse> searchAppointments( @Valid @RequestBody AppointmentQuery searchQuery) throws IOException {
+        if(searchQuery.getIsDatelessAppointments()) {
```
angshu, line 63: Also you are expecting a searchquery but fetching only date-less appointment? 
Does it mean 
- if appointment without dates, rest of criteria do not apply? That does not seem right. I can look for an appointment with a certain priority or service for which dates have not been assinged yet - right?

please don't call service method as "searchDatelessAppointment".  Call it "appointmentWithoutDatesAssigned" or something that makes more sense. DatelessAppt seems very much a technical internal thing.

omod/src/test/java/org/openmrs/module/appointments/web/mapper/AppointmentMapperTest.java
angshu: See comments above

angshu (changes requested):

gsluthra:
> > What is the default priority assumed, if no priority is set (for backward compatibility)? Where is that code? I am guessing default is "routine" priority. thanks!
> 
> Currently, default is not set, because showing it in the UI itself is behind a toggle. https://github.com/Bahmni/openmrs-module-appointments/pull/145/files#diff-0c4429900b456f54d785391170847fccd1f3e8f0fcb322d5ee296670318c2affR111 It'll be NULL only if we didn't pass any priority.

I am trying to determine, if an existing adopter of Bahmni, that already has appointments in their database, uses this new module, with wait-list enabled, how will previously created appointments show up (if they have null in priority). Is that something we need to worry, or the new feature/UI will work fine, if it has older created appointments (for future dates)?

api/src/test/java/org/openmrs/module/appointments/dao/impl/AppointmentDaoImplIT.java
```diff
@@ -243,27 +239,13 @@ public void shouldReturnAppointmentsForProvider() {
         assertEquals(1, appointments.size());
     }
 
-    @Test
-    public void shouldReturnAppointmentsForPatientBetweenGivenDatesForASpecificStatus() throws ParseException {
```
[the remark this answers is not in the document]
kavitha-sundararajan: Apologies, since it has been added later, it was removed while cherry picking by mistake. Added it back

api/src/test/java/org/openmrs/module/appointments/dao/impl/AppointmentDaoImplIT.java
kavitha-sundararajan: added it back

api/src/main/java/org/openmrs/module/appointments/dao/impl/AppointmentDaoImpl.java
```diff
@@ -160,13 +161,20 @@ public List<Appointment> search(AppointmentSearchRequest appointmentSearchReques
         Criteria criteria = sessionFactory.getCurrentSession().createCriteria(Appointment.class);
 
         criteria.add(Restrictions.eq("voided", false));
-        criteria.addOrder(Order.asc("startDateTime"));
+        if (appointmentSearchRequest.getStartDate() != null) {
+            criteria.addOrder(Order.asc("startDateTime"));
+        } else {
+            criteria.addOrder(Order.asc("dateCreated"));
```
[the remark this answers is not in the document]
kavitha-sundararajan: refactored to combine both criteria

api/src/main/java/org/openmrs/module/appointments/dao/impl/AppointmentDaoImpl.java
```diff
@@ -160,13 +161,20 @@ public List<Appointment> search(AppointmentSearchRequest appointmentSearchReques
...
+        }
         setDateCriteria(appointmentSearchRequest, criteria);
         setPatientCriteria(appointmentSearchRequest, criteria);
         setLimitCriteria(appointmentSearchRequest, criteria);
         setProviderCriteria(appointmentSearchRequest, criteria);
         setStatusCriteria(appointmentSearchRequest, criteria);
 
+        String limit = Context.getAdministrationService().getGlobalProperty("webservices.rest.maxResultsDefault");
```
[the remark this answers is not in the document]
kavitha-sundararajan: moved it to service method

omod/src/main/java/org/openmrs/module/appointments/web/contract/AppointmentQuery.java
```diff
@@ -12,6 +12,7 @@ public class AppointmentQuery {
     private String patientUuid;
     private String status;
     private String appointmentKind;
+    private Boolean isDatelessAppointments = false;
```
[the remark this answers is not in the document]
kavitha-sundararajan: renamed to withoutDates

omod/src/main/java/org/openmrs/module/appointments/web/controller/AppointmentController.java
kavitha-sundararajan: renamed getter as isWithoutDates() , service method as searchAppointmentsWithoutDates() and dao method as getAppointmentsWithoutDates()

api/src/main/resources/liquibase.xml
kavitha-sundararajan: default value is set to NULL now

api/src/main/java/org/openmrs/module/appointments/model/AppointmentPriority.java
```diff
@@ -0,0 +1,13 @@
+package org.openmrs.module.appointments.model;
+
+public enum AppointmentPriority {
+    AsNeeded("AsNeeded"), Routine("Routine"), Emergency("Emergency"), Invalid("Invalid");
```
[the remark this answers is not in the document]
kavitha-sundararajan: 1. Apologies, Invalid priority is not in use and we removed it later. Updated the PR
2. For changing the priority codes as of FHIR model now would require some more time since it might need changes from the frontend as well. So we created a card in the backlog to track it [BAH-3637](https://bahmni.atlassian.net/browse/BAH-3637) and will be taken up later.

omod/src/main/java/org/openmrs/module/appointments/web/contract/AppointmentRequest.java
kavitha-sundararajan: If we are to change it from String to Enum now, it would require some more time and we need to change it accordingly to the FHIR codes of priority. So we created a card in the backlog to track it [BAH-3637](https://bahmni.atlassian.net/browse/BAH-3637) and will be taken up later.

api/src/main/java/org/openmrs/module/appointments/dao/impl/AppointmentDaoImpl.java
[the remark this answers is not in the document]
angshu: I am little confused. We seem to be ordering by startdate (ascending) if its part of the search parameter. I can still want to search by status but order by startdate, no? If you want to sort based on request, then the request must send it as sort parameter, not search parameter

api/src/main/java/org/openmrs/module/appointments/helper/AppointmentServiceHelper.java
```diff
@@ -51,10 +51,14 @@ public String getAppointmentAsJsonString(Appointment appointment) throws IOExcep
         appointmentJson.put("providerUuid", providerUuid);
         String locationUuid = appointment.getLocation() != null ? appointment.getLocation().getUuid() : null;
         appointmentJson.put("locationUuid", locationUuid);
-        appointmentJson.put("startDateTime", appointment.getStartDateTime().toInstant().toString());
-        appointmentJson.put("endDateTime", appointment.getEndDateTime().toInstant().toString());
+        String startDate = (appointment.getStatus() == AppointmentStatus.WaitList && appointment.getStartDateTime() == null) ? null : appointment.getStartDateTime().toInstant().toString();
+        appointmentJson.put("startDateTime", startDate);
+        String endDate = (appointment.getStatus() == AppointmentStatus.WaitList && appointment.getStartDateTime() == null) ? null : appointment.getEndDateTime().toInstant().toString();
```
angshu, line 56: this will negate one of the criterias that a waitlisted appointment can have a date. This assumes that waitlist status appointment can not have a date. Why do we have to do this check against waitlist status. Why not simply just check if the dates are null, alternatively just get the time instant from the date? 

I am assuming the method "getAppointmentAsJsonString" is used for audit purpose, and since it has the appointment object - you have all the info.

omod/src/main/java/org/openmrs/module/appointments/web/contract/AppointmentQuery.java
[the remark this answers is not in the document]
angshu: "isDatelessAppointments" attribute still exists. Have you refactored the code to search by status instead of relying on "datelessAppointment" as part of search?

omod/src/main/java/org/openmrs/module/appointments/web/contract/AppointmentQuery.java
```diff
@@ -68,4 +69,12 @@ public String getServiceTypeUuid() {
     public void setServiceTypeUuid(String serviceTypeUuid) {
         this.serviceTypeUuid = serviceTypeUuid;
     }
+
+    public Boolean getIsDatelessAppointments() {
```
angshu: Why do we need this attribute "isDatelessAppointment"? if your idea is to search for waitlisted appointment - do it by "status" not this unnecessary attribute. Unless within a particular status, you also want to search the appointments without dates, which I believe is not your requirement.

omod/src/main/java/org/openmrs/module/appointments/web/controller/AppointmentController.java
angshu: same question above is relevant here as well. Why not leverage status if you are looking for "waitlisted" appointments? Unless you want to return all appointments without dates?

omod/src/main/java/org/openmrs/module/appointments/web/contract/AppointmentQuery.java
kavitha-sundararajan: this attribute was renamed to "withoutDates", not sure how it is showing "isDatelessAppointment". can we try refreshing the page once ?

api/src/main/java/org/openmrs/module/appointments/helper/AppointmentServiceHelper.java
kavitha-sundararajan: yeah, this method is used for audit purpose only. This will still allow waitlist appointments with dates by falling to the else part but there is no need of this check against waitlist. I'll remove the status check here

api/src/main/java/org/openmrs/module/appointments/dao/impl/AppointmentDaoImpl.java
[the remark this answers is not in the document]
kavitha-sundararajan: yeah I agree. reverted it back to the existing way since that'll still sort the appointments irrespective of status

angshu (approved):

kavitha-sundararajan (commented):
Few comments to be implemented are being tracked in this card [BAH-3637](https://bahmni.atlassian.net/browse/BAH-3637) and will be taken up later

PR #149, opened 2024-04-30
BAH-3239: Link appointments to the visits that fulfill them

mogoodrich:


api/src/main/java/org/openmrs/module/appointments/model/Appointment.java
```diff
@@ -250,5 +254,13 @@ public Date getDateCreated() {
     public void setDateCreated(Date dateCreated) {
         this.dateCreated = dateCreated;
     }
+
+    public Set<Encounter> getEncounters() {
+        return encounters;
```
mogoodrich, line 259: we could call this "fulfillingEncounters" to clarify that this meant to be used for encounters that fulfill an appointment (not encounters that *schedule* a specific appointment) but I think it's simpler just to use appointment)

api/src/main/resources/liquibase.xml
```diff
@@ -712,4 +712,22 @@
...
         </sql>
     </changeSet>
+    <changeSet id="add-patient-appointment-encounter-map-table-20240429" author="mgoodrich">
+        <preConditions onFail="MARK_RAN">
+            <not>
+                <tableExists tableName="patient_appointment_encounter_map" />
+            </not>
+        </preConditions>
+        <comment>Add the map table to link appointments to encounter</comment>
+        <sql>
```
mseaton, line 722: We should not use raw sql for this, especially not when it contains backticks, which seems very mysql-specific.  Liquibase has a <createTable...> element, with correponding <column> elements, as well as a <addForeignKeyContstraint> element.  These should be used preferably.  There is also no need for the `IF NOT EXISTS` statement given the precondition that is on the overall changeset.

api/src/main/resources/Appointment.hbm.xml
```diff
@@ -62,6 +62,13 @@
             <one-to-many class="AppointmentAudit"/>
         </set>
 
+        <set name="encounters" table="patient_appointment_encounter_map" lazy="true" cascade="none">
+            <key column="patient_appointment_id" />
+            <many-to-many class="org.openmrs.Encounter">
+                <column name="encounter_id" />
+            </many-to-many>
+        </set>
```
mseaton, line 70: I think we need to be really careful here, and ensure our tests cover the various cascading cases that exist, both from the Hibernate perspective as well as the OpenMRS saveHandlers perspective.  i.e. we don't necessarily want to to find that voiding an Appointment ends up voiding all of the associated Encounters.  Or that deleting it from the database deletes the Encounter rows as well.

api/src/main/java/org/openmrs/module/appointments/model/Appointment.java
mseaton: Personally, I prefer `fulfillingEncounters` or `fulfillmentEncounters`.  As you say, otherwise there is ambiguity - could these encounters include the ones that scheduled the appointment?  If the intent is solely to represent Encounters that fulfill the appointment (and I believe it is), then we should be explicit.

api/src/main/resources/liquibase.xml
```diff
@@ -712,4 +712,22 @@
             ALTER TABLE patient_appointment MODIFY end_date_time datetime default null;
         </sql>
     </changeSet>
+    <changeSet id="add-patient-appointment-encounter-map-table-20240429" author="mgoodrich">
+        <preConditions onFail="MARK_RAN">
+            <not>
+                <tableExists tableName="patient_appointment_encounter_map" />
```
mseaton, line 718: What about naming this `patient_appointment_fulfillment_encounter`?

mseaton (changes requested):

api/src/main/resources/liquibase.xml
mogoodrich: Fair enough

api/src/main/resources/Appointment.hbm.xml
mogoodrich: Fair enough, I did set cascade=none, but good to actually test what this does, I will add some tests.

api/src/main/resources/liquibase.xml
mogoodrich: The OpenMRS convention seems to be end these tables with "map" (but, yes, if we change the name of the parameter to fulfillment or fulfilling, we should add that to this table.

api/src/main/java/org/openmrs/module/appointments/model/Appointment.java
```diff
@@ -250,5 +254,13 @@ public Date getDateCreated() {
...
         this.dateCreated = dateCreated;
     }
+
+    public Set<Encounter> getFulfillingEncounters() {
+        return fulfillingEncounters;
+    }
+
+    public void setFulfillingEncounters(Set<Encounter> fulfillingEncounters) {
+        this.fulfillingEncounters = fulfillingEncounters;
+    }
```
mogoodrich, line 264: Changed to "fulfillingEncounters" per code review feedback.

api/src/test/java/org/openmrs/module/appointments/dao/impl/AppointmentDaoImplIT.java
```diff
@@ -306,4 +316,45 @@ public void shouldReturnAppointmentsWithoutDates() {
...
+        Context.flushSession();
+        Context.clearSession();
+
+        appointment = appointmentDao.getAppointmentByUuid(appointmentUuid);
+        assertTrue(appointment.getVoided());
+        Encounter enc1 = encounterService.getEncounterByUuid("f303e49f-24a9-41bb-810b-5d07b881c0e0");
+        assertFalse(enc1.getVoided());
+        Encounter enc2 = encounterService.getEncounterByUuid("785826c1-f137-4b4f-bb2f-07b5c546d139");
+        assertFalse(enc2.getVoided());
+    }
```
mogoodrich, line 343: So, after writing this, I don't know if this (or the new service test class and method I added) really test much, as I think the save handlers and void handlers depend on the Appointments Service being an "OpenMRS Service", which the Appointments Service is not. Also, there is no "voidAppointment" method at the API or DB level.  So I don't think there's much risk of voided getting cascaded down to the encounter, but hopefully (though probably doubtful) one of these new voiding tests will catch a problem if it gets introduced in the future.

mseaton (approved):
Looks good to me.  Hopefully we can get someone from Bahmni to weigh in? @angshu ?  @gsluthra ?

mogoodrich:
Thanks @mseaton !  @angshu , @gsluthra, thoughts?  It would be great to get this merged in as we need it for some functionality we are planning on building.

api/src/main/resources/liquibase.xml
```diff
@@ -712,4 +712,23 @@
...
         </sql>
     </changeSet>
+    <changeSet id="add-patient-appointment-fulfilling-encounter-map-table-20240501" author="mgoodrich">
+        <preConditions onFail="MARK_RAN">
+            <not>
+                <tableExists tableName="patient_appointment_fulfilling_encounter_map" />
+            </not>
+        </preConditions>
+        <comment>Add the map table to link appointments to fulfilling encounter</comment>
+        <createTable tableName="patient_appointment_fulfilling_encounter_map">
```
angshu, line 722: should we have "dateCreated", "createdBy" columns as well?

api/src/test/java/org/openmrs/module/appointments/dao/impl/AppointmentDaoImplIT.java
angshu: do we see a case, when an appointment (post fulfilment through an encounter gets) gets voided? Should this be allowed? If not, should we put check/validate and restrict voiding of appointment in such case?  if we do allow that - then should we remove the association with the encounter?

api/src/main/resources/liquibase.xml
mseaton: It's a good question @angshu .  My original take on this was that this is not really an independent domain object, but more to support a Collection property on Encounter and Appointment so we don't really need them.  i.e. this is essentially here to support both a Set<Encounter> property on appointment, and a (virtual) Set<Appointment> property on Encounter.  But if the typical pattern is to include Auditable properties on objects like this, or if there are known use cases that require it, then we can do that.

mogoodrich: Yeah, @mseaton  and I went back and forth on this a bit and decided that having auditable fields on this table probably wasn't needed for the reasons Mike suggests.

I'd be inclined to keep it is unless you feel strongly it's needed @angshu ?  We'd also need to build in proper cascading voidng (ie when you void the appointment you should void the appointment_encounter_map), and we'd want to build full void support into the Appointments Service, as that doesn't appear to be 100% implemented now.

angshu: no, I am ok. thank you

angshu (approved):

mogoodrich:
Thanks @angshu !

PR #151, opened 2024-05-07
Kavitha|Awaiting appointments code refactor

kavitha-sundararajan:


api/src/main/java/org/openmrs/module/appointments/service/impl/AppointmentsServiceImpl.java
```diff
@@ -29,14 +30,7 @@
...
 import java.time.Instant;
-import java.util.Collections;
-import java.util.Date;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Optional;
-import java.util.Set;
+import java.util.*;
```
Arjun-Go, line 33: Can we have specific imports rather than * import

api/src/test/java/org/openmrs/module/appointments/service/impl/AppointmentsServiceImplTest.java
```diff
@@ -5,18 +5,13 @@
...
-import org.mockito.ArgumentCaptor;
-import org.mockito.InjectMocks;
-import org.mockito.Mock;
-import org.mockito.MockitoAnnotations;
-import org.mockito.Spy;
-import org.openmrs.Patient;
-import org.openmrs.Person;
-import org.openmrs.Provider;
-import org.openmrs.User;
+import org.mockito.*;
```
Arjun-Go, line 8: Same here

Arjun-Go (approved):

api/src/main/java/org/openmrs/module/appointments/service/impl/AppointmentsServiceImpl.java
kavitha-sundararajan: done

api/src/test/java/org/openmrs/module/appointments/service/impl/AppointmentsServiceImplTest.java
kavitha-sundararajan: done

PR #162, opened 2024-10-03
PLT-82: Add PostgreSQL support to reference application modules

wikumChamith:
Fixes PostgreSQL compatibility issues in the liquibase.xml

### DIFF
--- a/api/src/main/resources/liquibase.xml
+++ b/api/src/main/resources/liquibase.xml
@@ -6,68 +6,130 @@
          http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-2.0.xsd">
 
     <changeSet id="create-appointment_speciality_table-201707031130" author="Shruthi,Pushpa">
+        <validCheckSum>8:22c4008a265e42718db0f7eccc7a27a6</validCheckSum>
         <preConditions onFail="MARK_RAN">
             <not>
                 <tableExists tableName="appointment_speciality" />
             </not>
         </preConditions>
 
-        <sql>
-            CREATE  TABLE IF NOT EXISTS appointment_speciality (
-            speciality_id INT NOT NULL AUTO_INCREMENT ,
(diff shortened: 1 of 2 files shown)


wikumChamith:
@angshu can you take a look at this?

rahu1ramesh (changes requested):

mherman22:
Hey @wikumChamith , so with these changes the Initializer complains with the following error: `Caused by: org.postgresql.util.PSQLException: ERROR: relation "speciality_id_seq" does not exist`

The error i believe occurred because the Hibernate mapping explicitly referenced a sequence named `speciality_id_seq`, but PostgreSQL did not have such a sequence. In PostgreSQL, when a column is defined using `autoIncrement="true"` in Liquibase, it translates to `GENERATED BY DEFAULT AS IDENTITY`, which creates a sequence with an auto-generated name (e.g., `appointment_speciality_speciality_id_seq`). Since no sequence named `speciality_id_seq` existed, PostgreSQL threw a `relation "speciality_id_seq" does not exist` error during entity persistence.

The fix was to remove the explicit sequence name from the Hibernate mapping and rely on the `native` ID generation strategy by changing:

```xml
<generator class="native">
    <param name="sequence">speciality_id_seq</param>
</generator>
```

to simply:

```xml
<generator class="native"/>
```

Also did the above step for https://github.com/Bahmni/openmrs-module-appointments/blob/e689087848c03b115c0675b1988b806657f744a9/api/src/main/resources/AppointmentServiceDefinition.hbm.xml#L8-L10 in order to fix `Caused by: org.postgresql.util.PSQLException: ERROR: relation "appointment_service_id_seq" does not exist`

Here are the logs just incase you need to look them over: https://paste.ubuntu.com/p/RYWwsD3ktS/

cc: @dkayiwa

angshu:
Just following up on this. @wikumChamith @mherman22 any resolution to this?

PR #163, opened 2024-11-28
Add indexes on patient appointment start_date_time and status columns…

ojwanganto:
This PR adds indexes on start_date_time and status columns for improved query perfornance

### DIFF
--- a/api/src/main/resources/liquibase.xml
+++ b/api/src/main/resources/liquibase.xml
@@ -749,4 +749,14 @@
             update patient_appointment set date_appointment_scheduled = date_created where date_appointment_scheduled is null;
         </sql>
     </changeSet>
+    <changeSet author="aojwang" id="add_index_on_patient_appointment_start_date_time_202411281139">
+        <createIndex indexName="patient_appointment_start_date_time_index" tableName="patient_appointment">
+            <column name="start_date_time"/>
+        </createIndex>
+    </changeSet>
+    <changeSet author="aojwang" id="add_index_on_patient_appointment_status_202411281152">
+        <createIndex indexName="patient_appointment_status_index" tableName="patient_appointment">
+            <column name="status"/>
+        </createIndex>
+    </changeSet>
 </databaseChangeLog>


api/src/main/resources/liquibase.xml
```diff
@@ -749,4 +749,14 @@
...
         </sql>
     </changeSet>
+    <changeSet author="aojwang" id="add_index_on_patient_appointment_start_date_time_202411281139">
+        <createIndex indexName="patient_appointment_start_date_time_index" tableName="patient_appointment">
+            <column name="start_date_time"/>
+        </createIndex>
+    </changeSet>
+    <changeSet author="aojwang" id="add_index_on_patient_appointment_status_202411281152">
+        <createIndex indexName="patient_appointment_status_index" tableName="patient_appointment">
+            <column name="status"/>
```
gsluthra, line 759: I am not sure if a status index is helpful, since a huge number of appointments, over time, will have the same status. If index on date-time is already there, then the set of appointments is already narrowed to a small set, where selecting status should not be much work for DB while querying.

Can you validate if the speed up is worth it, with date_time vs (date_time+status)?

ojwanganto: I have conducted a few tests and your POV is accurate. I ran queries with and without the index on status and the execution time is the same.

ojwanganto:
I have updated the PR. Please re-review @gsluthra

ojwanganto:
ping @gsluthra @angshu

api/src/main/resources/liquibase.xml
```diff
@@ -749,4 +749,10 @@
             update patient_appointment set date_appointment_scheduled = date_created where date_appointment_scheduled is null;
         </sql>
     </changeSet>
+    <changeSet author="bahmni-appointments" id="add_index_on_patient_appointment_start_date_time_202411281139">
+        <comment>add index on patient_appointment start_date_time column</comment>
+        <createIndex indexName="patient_appointment_start_date_time_index" tableName="patient_appointment">
```
rahu1ramesh, line 754: @ojwanganto Thank you for the update! The changeset correctly adds an index to the `start_date_time` column in the `patient_appointment` table. However, to **prevent duplicate index creation errors**, it would be best to include a **precondition** to check if the index already exists before creating it.  

Please add the following **precondition** before the `<createIndex>` tag:  

```xml
<preConditions onFail="MARK_RAN">
    <not>
        <indexExists indexName="patient_appointment_start_date_time_index" tableName="patient_appointment"/>
    </not>
</preConditions>
```

Adding a precondition would prevents redundant index creation, help avoid migration failures if the index already exists and will ensure safe and idempotent schema updates.

Could you please update the changeset accordingly? Let me know if you have any questions. Thanks!

rahu1ramesh (changes requested):

PR #195, opened 2026-04-14
BAH-4637 - Add an api to change the status of list of appointments

PreethiKalyani99:


api/src/main/java/org/openmrs/module/appointments/service/impl/AppointmentsServiceImpl.java
```diff
@@ -467,4 +467,31 @@ private void createAndSetAppointmentAudit(Appointment appointment) {
...
+    @Transactional
+    @Override
+    public List<Appointment> changeStatusForAppointments(List<String> appointmentUuids, String toStatus) {
+        log.info("Changing status for " + appointmentUuids.size() + " appointment(s) to: " + toStatus);
+
+        List<Appointment> appointments = appointmentDao.getAppointmentsByUuids(appointmentUuids);
+
+        if (appointments.isEmpty()) {
+            log.error("No valid appointments found for the provided UUIDs");
+            throw new APIException("No valid appointments found for the provided UUIDs");
```
angshu: Should it not be illegalArgument as well?

api/src/main/java/org/openmrs/module/appointments/service/AppointmentsService.java
```diff
@@ -109,5 +109,9 @@ public interface AppointmentsService {
     @Transactional
     @Authorized({VIEW_APPOINTMENTS, MANAGE_APPOINTMENTS})
     List<Appointment> searchAppointmentsWithoutDates(AppointmentSearchRequestModel searchQuery);
+
+    @Transactional
+    @Authorized({MANAGE_APPOINTMENTS, MANAGE_OWN_APPOINTMENTS})
+    List<Appointment> changeStatusForAppointments(List<String> appointmentUuids, String toStatus);
```
angshu: Why not take the status as enum? On the controller, you are taking in as ENUM in the request. 
I understand that the changeStatus() method takes in a string .. but this API can correct that.

omod/src/main/java/org/openmrs/module/appointments/web/contract/AppointmentStatusChangeRequest.java
```diff
@@ -0,0 +1,33 @@
...
+
+import org.openmrs.module.appointments.model.AppointmentStatus;
+
+import javax.validation.constraints.NotNull;
+import javax.validation.constraints.Size;
+import java.util.List;
+
+public class AppointmentStatusChangeRequest {
+    
+    @NotNull(message = "appointmentUuids array is required in request body")
```
angshu: better message maybe?
1. list of appointment uuid is required
2. at least one appointment uuid must be specified
3. toStatus is required

angshu (approved):

PR #200, opened 2026-06-30
Add appointment unavailability conflict - checks unavailability at dateAndTime, location, service and provider level

PreethiKalyani99:
## Summary by CodeRabbit

* **New Features**
  * Added support for detecting appointment unavailability conflicts.
  * Appointment conflict checks now respect a location’s time zone when available.

* **Bug Fixes**
  * Improved time-based conflict matching so appointments are compared using the correct local day and time.
  * Added fallback behavior when no location time zone is configured.
  * Ensured unavailability rules can match by location, service, and provider as expected.

* **Tests**
  * Added and updated tests covering time zone handling and unavailability conflict scenarios.

### DIFF
--- a/api/src/main/java/org/openmrs/module/appointments/conflicts/impl/AppointmentServiceUnavailabilityConflict.java
+++ b/api/src/main/java/org/openmrs/module/appointments/conflicts/impl/AppointmentServiceUnavailabilityConflict.java
@@ -1,19 +1,19 @@
 package org.openmrs.module.appointments.conflicts.impl;
 
 import org.openmrs.module.appointments.conflicts.AppointmentConflict;
-import org.openmrs.module.appointments.model.AppointmentConflictType;
 import org.openmrs.module.appointments.model.Appointment;
+import org.openmrs.module.appointments.model.AppointmentConflictType;
 import org.openmrs.module.appointments.model.AppointmentServiceDefinition;
 import org.openmrs.module.appointments.model.ServiceWeeklyAvailability;
 import org.openmrs.module.appointments.util.DateUtil;
 
 import java.sql.Time;
 import java.text.SimpleDateFormat;
+import java.time.ZoneId;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.List;
 import java.util.Objects;
-import java.util.Optional;
 import java.util.Set;
 import java.util.stream.Collectors;
 
@@ -22,7 +22,7 @@ import static org.openmrs.module.appointments.util.DateUtil.getEpochTime;
 
 public class AppointmentServiceUnavailabilityConflict implements AppointmentConflict {
 
-    private final static String DAY_OF_WEEK_PATTERN = "EEEE";
+    private static final String DAY_OF_WEEK_PATTERN = "EEEE";
     private final SimpleDateFormat DayFormat = new SimpleDateFormat(DAY_OF_WEEK_PATTERN);
 
     @Override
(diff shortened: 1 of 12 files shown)


api/src/main/java/org/openmrs/module/appointments/util/DateUtil.java
```diff
@@ -76,6 +83,26 @@ public static long getEpochTime(long date) {
...
+
+    public static Optional<ZoneId> getLocationZone(Location location) {
+        if (location == null) return Optional.empty();
+        return location.getActiveAttributes().stream()
+                .filter(attr -> TIMEZONE_ATTRIBUTE.equalsIgnoreCase(attr.getAttributeType().getName()))
+                .map(attr -> {
+                    try {
+                        return ZoneId.of(String.valueOf(attr.getValue()));
+                    } catch (DateTimeException e) {
+                        return null;
```
mohan-13, line 99: its okay to return null, but good to have a log.error so that debugging becomes easire

mohan-13: Mixed patterns here, the first block returns Optional.empty whereas here a null is returned. Potential for NullPointerException

api/src/main/java/org/openmrs/module/appointments/service/impl/AppointmentUnavailabilityServiceImpl.java
```diff
@@ -29,12 +31,13 @@ public class AppointmentUnavailabilityServiceImpl implements AppointmentUnavaila
     private Log log = LogFactory.getLog(this.getClass());
 
     private final AppointmentUnavailabilityDao appointmentUnavailabilityDao;
-    private final AppointmentServiceDefinitionService appointmentServiceDefinitionService;
 
-    public AppointmentUnavailabilityServiceImpl(AppointmentUnavailabilityDao appointmentUnavailabilityDao,
-                                                 AppointmentServiceDefinitionService appointmentServiceDefinitionService) {
+    @Autowired
```
mohan-13, line 35: Why is this marked as Autowired ? this should have been a bean already in moduleApplicationContext.xml

api/src/main/java/org/openmrs/module/appointments/conflicts/impl/AppointmentUnavailabilityConflict.java
```diff
@@ -0,0 +1,86 @@
...
+
+    @Override
+    public List<Appointment> getConflicts(List<Appointment> appointments) {
+        return appointments.stream()
+                .filter(this::hasConflict)
+                .collect(Collectors.toList());
+    }
+
+    private boolean hasConflict(Appointment appointment) {
+        ZoneId zone = DateUtil.getLocationZone(appointment.getLocation()).orElse(ZoneId.systemDefault());
```
mohan-13, line 44: different handling in multiple places. The above conflict sets ZoneId as null, where as here it is system default. Can we extract out methods so that it is similar across different handling.

api/src/main/java/org/openmrs/module/appointments/conflicts/impl/AppointmentServiceUnavailabilityConflict.java
```diff
@@ -42,36 +42,42 @@ public List<Appointment> getConflicts(List<Appointment> appointments) {
         return conflictingAppointments;
     }
 
-    private boolean checkConflicts(Appointment appointment, AppointmentServiceDefinition appointmentServiceDefinition) {
-        Set<ServiceWeeklyAvailability> weeklyAvailableDays = appointmentServiceDefinition.getWeeklyAvailability();
+    private boolean checkConflicts(Appointment appointment, AppointmentServiceDefinition service) {
+        ZoneId zone = DateUtil.getLocationZone(appointment.getLocation()).orElse(null);
```
mohan-13, line 46: As discussed, we just don't rely on the appointment.location(), we should also check if the appointment service has a location. If it haswe should consider that as well

PR #201, opened 2026-07-01
BAH-4840: Add date/time, location, service, and provider conflict checks for ap…

PreethiKalyani99:
## Ticket 

https://bahmni.atlassian.net/browse/BAH-4840

## Description

**Unavailability period conflict detection:**
Previously, AppointmentServiceUnavailabilityConflict only checked whether an appointment fell outside the service's defined weekly hours. 
Appointments are now validated against AppointmentUnavailability records. The conflict is evaluated at multiple levels — date/time, location, service, and provider. If an unavailability block is defined for a specific location, service, or provider, it only conflicts with matching appointments. If no provider is specified, only the date/time, location, and service are used to determine whether a conflict exists.

**Timezone-aware service hours comparison:**
Service hours (weekly availability slots and start/end times) are now compared using the timezone derived from the appointment's location. If the appointment's location has no timezone configured, the service's location timezone is used. If neither is available, the system default is applied. 
Previously, all service hours comparisons used the system default timezone regardless of where the appointment was scheduled, which could produce incorrect conflict results for appointments in a different timezone than the server.

**Overnight and multi-day unavailability overlap:**
The overlap check previously worked by first verifying that the appointment date fell within the unavailability date range, then separately checking whether the appointment's clock times overlapped with the unavailability's clock times. This two-step approach broke down for unavailability periods that span midnight or multiple days, because the start and end times belong to different calendar dates. For example, an unavailability from 28-Jun 18:00 to 29-Jun 09:00 — when checked against a 29-Jun 08:00 appointment — would compare 08:30 against 18:00 as bare clock times and conclude there is no overlap, even though the appointment clearly falls within the blocked period.

The fix replaces this with a single `ZonedDateTime` interval comparison: the unavailability's start date/time and end date/time are each constructed as a full timestamp in the appointment's timezone, and the check becomes `apptStart.isBefore(unavailEnd) && apptEnd.isAfter(unavailStart)`. This is the standard interval overlap formula and correctly handles same-day, overnight, and multi-day unavailability periods without any special casing.

### DIFF
--- a/api/src/test/java/org/openmrs/module/appointments/conflicts/impl/AppointmentServiceUnavailabilityConflictTest.java
+++ b/api/src/test/java/org/openmrs/module/appointments/conflicts/impl/AppointmentServiceUnavailabilityConflictTest.java
@@ -414,4 +414,21 @@ public class AppointmentServiceUnavailabilityConflictTest {
         assertEquals(1, conflicts.size());
         assertEquals(appointment, conflicts.get(0));
     }
+
+    @Test
+    public void shouldReturnConflictWhenAppointmentFallsWithinOvernightUnavailability() {
+        AppointmentUnavailability unavailability = buildUnavailability("2026-06-28", "18:00:00", "2026-06-29", "09:00:00");
+        when(appointmentUnavailabilityDao.getAll(any(AppointmentUnavailabilitySearchParams.class)))
+                .thenReturn(Collections.singletonList(unavailability));
+
+        Appointment appointment = new Appointment();
+        appointment.setStartDateTime(getDate(2026, 5, 29, 8, 0, 0));
+        appointment.setEndDateTime(getDate(2026, 5, 29, 8, 30, 0));
+
+        List<Appointment> conflicts = appointmentServiceUnavailabilityConflict.getConflicts(Collections.singletonList(appointment));
+
+        assertNotNull(conflicts);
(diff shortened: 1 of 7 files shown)


api/src/test/java/org/openmrs/module/appointments/conflicts/impl/AppointmentServiceUnavailabilityConflictTest.java
```diff
@@ -27,11 +42,48 @@ public class AppointmentServiceUnavailabilityConflictTest {
...
+        appointment.setEndDateTime(getDate(2026, 5, 28, 20, 0, 0));
+        return appointment;
+    }
+
     @Test
     public void shouldReturnServiceUnavailableDayConflicts() {
         AppointmentServiceDefinition appointmentServiceDefinition = new AppointmentServiceDefinition();
         Appointment appointment = new Appointment();
         appointment.setService(appointmentServiceDefinition);
+        appointment.setLocation(createLocationWithSystemTimezone());
```
mohan-13: Why is the location being added to existing tests ? The expectation is that existing configuration should work even if location is null in the appointment request

mohan-13:
Please add JIRA card reference and update the details. Also please add a docs to the PR so that the flow can be understood

mohan-13 (approved):

